PR #14, opened 2024-01-29
Feature/episodes

### LINKED ISSUE #8: Abstraction hierarchy: actions - episodes - sagas/case studies
The team felt there should be a layer in between individual (or 'atomic') actions performed on individual letters ('writing', 'delivering' etc.) and the overall grouping of all data in case studies (or 'sagas').

Actions should be grouped in a new category called 'Episodes' (vel sim.), which should in turn be grouped into 'Case Studies'.

XanderVertegaal:
This is a starting point for a discussion on how to resolve #8 

In this PR:
- The old 'Epistolary Event' model has been split in two models:
  - Letter Actions are atomic uses of a letter. 
  - Epistolary Events are collections of letter actions.
- A new model 'Case Study' has been added.
- More human-friendly texts are shown in the admin forms. For instance: the case study form now says 'Add another Epistolary event' instead of the default "Add another Casestudy_epistolaryevent relationship".


backend/data/admin.py
```diff
@@ -4,64 +4,112 @@
 
 class PersonNameAdmin(admin.StackedInline):
     model = models.PersonName
-    fields = ['value', 'certainty', 'note']
-
+    fields = ["value", "certainty", "note"]
```
XanderVertegaal, line 7: My editor automatically uses Black. I could turn this off for this project, but it may not be a bad idea to agree on a formatter.

backend/data/models.py
```diff
@@ -69,76 +70,131 @@ class PersonName(Field, models.Model):
...
+
+
+class LetterAction(models.Model):
+    """
+    A letter action is an atomic action performed on a letter, e.g. writing, delivering, reading.
+
+    These can be grouped into epistolary events.
+    """
+
     letters = models.ManyToManyField(
```
XanderVertegaal, line 136: A `LetterAction` currently contains multiple letters; it inherited this from the old `EpistolaryEvent` model. What would you say if we made this a 1-to-1 relationship? That would fit the idea of an 'atomic' action more.

backend/data/models.py
```diff
@@ -69,76 +70,131 @@ class PersonName(Field, models.Model):
...
+        return self.name
+
+
 class EpistolaryEvent(models.Model):
+    """
+    Epistolary events are groups of related letter actions that are connected in some way.
+
+    For instance, a political campaign (epistolary event) may consist of the writing, transporting and reading of individual letters (letter actions).
+    """
+
```
XanderVertegaal, line 106: I considered making EpistolaryEvents nestable, so one epistolary event (e.g. 'the Election of Pope Clement III') can have multiple Epistolary Events (e.g. 'Council of Troyes' and 'Murder of Baldwin of Nantes') etc., but we should probably ask the team first before we needlessly add more layers of complexity.

backend/data/admin.py
lukavdplas: I personally don't mind formatting inconsistencies, but if you prefer we use the same formatter, I can install it.

lukavdplas: What I _do_ mind is (auto-)formatting an entire file when you're changing a few lines of content. It makes the review more difficult, as meaningful changes get cluttered by meaningless ones.

Anyway, we can choose something in our next meeting. It's a good idea to decide on this early in the project.

backend/data/models.py
lukavdplas: I had the same idea. Indeed, let's do it like this for now, and check how the team wants to connect events. I can also imagine horizontal connections rather than hierarchical ones.

backend/data/models.py
lukavdplas: I would agree that "atomic" implies it's one-to-one, but maybe this doesn't make sense for the researchers. If a messenger delivers three letters at once, you could consider that an "atomic action" as it can't be broken up into a sequence of actions.

I'd keep it like this for now, and check in in our next meeting.

lukavdplas (approved):
Nice!

backend/data/admin.py
XanderVertegaal: Agreed! In the meantime, I'll turn off Black for this project to avoid large meaningless diffs.

PR #16, opened 2024-01-31
Feature/separate apps

### LINKED ISSUE #15: Separate backend into apps
Separate the backend into apps to have smaller, more maintainable files.

E.g. 

- `core`/`shared`
- `person`
- `letter`
- `event`

lukavdplas:
This separates the backend into several modules (close #15)

> [!WARNING]
> Running these migrations will *remove all data* from the database, with the exception of user/authentication data. No workaround is implemented because of the current stage in the project.


XanderVertegaal (approved):
Nice work!

This should make it a lot easier to find specific models or admin configurations.

Issue #3, opened 2023-11-16
Set up deployed environment


lukavdplas:
- Deploy the application
- Set up group permissions for researchers
- Add user accounts

PR #17, opened 2024-02-01
Feature/letter model updates

### LINKED ISSUE #10: Model changes for Letter
- Change representation in Django Admin. Instead of "letter #57" there should be a name / one-line identifier so the team can quickly find/identify new letters, e.g. ('No. 60: Boniface to pope Zacharias in 578')
- Add `sender`
- Add `addressee`
- Add `date`
- Add `type` (e.g. 'friendship letter', 'consensus letter', 'royal order'). This could be an editable category in its own right?
- Instead of `certainty`, two new fields: `historically_attested` (letter features in actual, historical events vs. letter features in story/legend/fairytale) and `mentioned`: (letter is directly mentioned vs. existence of letter is implied/possible).

lukavdplas:
Expands the data model for letters (close #10)

- Adds a `name` field to letters, which is required and unique, and used as the string representation. It will default to the current `letter #3` string representation during the migration, though if you run this right after #16, this doesn't really matter. Note that this operation requires 3 (!) migration scripts, following [django documentation on adding unique non-nullable fields](https://docs.djangoproject.com/en/5.0/howto/writing-migrations/#migrations-that-add-unique-fields).
- Adds a `category` field to letters. Letter categories are a model by themselves, so researchers can add or remove categories as needed.
- Adds `date_active` and `date_written` readonly fields to the admin interface. The values of these fields is based on the actions associated with the letter. They give the date range of, respectively, all actions involving the letters, and the "writing" actions involving the letter.
- Adds `addressees` and `senders` fields to letters.


backend/letter/models.py
```diff
@@ -1,10 +1,62 @@
...
+        return self.name
+
+    @admin.display(
+        description="Date range of actions involving this letter",
+    )
+    def date_active(self):
+        return self._aggregate_dates(self.events.all())
+
+    @admin.display(
+        description="Date range in which this letter was written",
```
XanderVertegaal, line 25: This is great! Good idea to add this!

XanderVertegaal (approved):
All good changes! I think Hope and Robert will be pleased now letters will have names that are more insightful than simply 'letter #4' vel sim.

(I'm glad you added the link to the Django docs about the three migration scripts needed to insert a unique non-nullable field. It would have been quite confusing otherwise. ;))

PR #20, opened 2024-02-02
Feature/person model updates

### LINKED ISSUE #11: Model changes for Person
* Add field `gender`
* Add field `social status`
* Add field `occupation/office`. This should be tied to a date (e.g. Boniface was only a bishop for part of his life).

I guess we might also add `date of birth` and `date of death`, allowing for the option `unknown` in case we don't have any clue.

### LINKED ISSUE #18: outfactor `DateField` class
The `LetterEventDate` contains the basic data format for entering dates and includes a relation to letter actions.

We'll need to enter dates in other contexts, so it will be handy a `DateField` class in the `core` app that defines the format, global date range. We could also add some utility methods to the class to aggregate or display dates. Fields like `LetterEventDate` can inherit this class.

XanderVertegaal:
This PR updates the Person model (resolves #11) and outfactors `LetterActionDate` to the `core` module (resolves #18).

**Note: this branch is based on feature/separate-apps, which should be merged first.**

In more detail:
- The `unique=True` on the `PersonName.person` field has been replaced by a `UniqueConstraint`. Adding multiple names to one person would not work properly before, but now it does. This also removes a recurring warning about having `ForeignKey(unique=True)`, which Django does not like. `Person.__str__()` has been updated appropriately to deal with multiple names.
- Tests have been added for the display of Person names, but I am not sure if the way I did it is the way to go. (See comments below; input is appreciated.)

- Added models: `Office` (fancy name for what is essentially a 'job') and `Occupation` (for the relationship between a `Person` and an `Office`), `PersonDateOfBirth` and `PersonDateOfDeath`.

- The `LetterActionDate` model has been outfactored to an abstract model called `LettercraftDate`. The following models subclass it: `LetterEventDate`, `PersonDateOfBirth`, `PersonDateOfDeath`, `Occupation`.

- Running `pytest` displayed warnings about a deprecated setting in `settings.py`. Since we do not use it and it does not have a function anymore, I removed it.


backend/conftest.py
```diff
@@ -44,3 +44,37 @@ def epistolary_event(db, letter, case_study):
...
+
+
+@pytest.fixture()
+def person_unnamed(db):
+    person = Person.objects.create()
+    return person
+
+
+@pytest.fixture()
+def person_single_name(db):
```
XanderVertegaal, line 56: In order to test the result of `person.__str__()` if a `Person` has no, one or multiple added names, I need to create several instances of Person somehow (or add names progressively).

I am now creating several fixtures, but I can see how this gets out of hand quite quickly. I am sure there is a more convenient/standard way of performing the "arrange" part of `pytest` tests. Is there a convention/example in the lab that I can follow?

backend/lettercraft/settings.py
```diff
@@ -120,8 +120,6 @@
 
 USE_I18N = True
 
-USE_L10N = True
```
XanderVertegaal, line 123: This setting produces warnings in `pytest` as it is deprecated since Django 4.0 (we are running 4.2). `True` is the default setting anyway, so I think we can safely remove it.

package.json
```diff
@@ -21,7 +21,7 @@
         "test-back": "yarn back pytest",
         "test-func": "yarn func pytest",
         "test": "yarn test-back && yarn test-front && yarn test-func",
-        "start-front": "yarn fyarn stop && yarn front yarn serve",
+        "start-front": "yarn front yarn serve",
```
XanderVertegaal, line 24: This is not really necessary, but it makes my life as a Windows user a lot easier. `fyarn stop` uses `lsof`, a Linux-only command.

backend/conftest.py
lukavdplas: We don't have any naming conventions, but here are some things I do in cases like this:

First, you can provide fixtures on different scopes. `/backend/conftest.py` is the highest scope for application-wide fixtures. In this case, you could provide these in `/backend/person/conftest.py` (for fixtures in the person app) or even in `/backend/person/tests/test_person_models.py` (i.e. right in the test file). It depends on where you think the fixtures could be reused. In this case, providing them directly in the test file would be appropriate.

Second, you don't _need_  a fixture if you're only using it once. In this case, you could write your test like so:

```py
def test_single_name(db):
    person = Person.objects.create()
    PersonName.objects.create(person=person, value="Bert")
    assert str(person) == "Bert"
```

That's a perfectly fine test. Making a fixture for a single test can help with readability if your test requires elaborate setup, but this one is pretty short. This can get repetitive between tests, however.

Finally, the state of the database is not preserved between tests unless you want it to be. In this case, you should be able to write your tests with fixtures like these:

```python
@pytest.fixture()
def person(db):
    return Person.objects.create()

@pytest.fixture()
def person_name(db, person):
    return PersonName.objects.create(person=person, name="Bert")

@pytest.fixture()
def person_name_2(db, person):
    return PersonName.objects.create(person=person, name="Ernie")

def test_person_no_name(person):
    # assert str(person) == ...

def test_person_single_name(person, person_name):
    # assert str(person) == ...

def test_person_multiple_name(person, person_name, person_name_2):
    # assert str(person) == ...

```

backend/person/models.py
```diff
@@ -1,28 +1,108 @@
...
+        max_length=256,
+        help_text="The name of the office (e.g. 'bishop of Toulouse', 'king of France')",
+    )
+    description = models.TextField(
+        blank=True,
+        help_text="A description of the office (e.g. 'The bishop of Toulouse is the ecclesiastical ruler of the diocese of Toulouse')",
+    )
+
+    def __str__(self):
+        return self.name
```
lukavdplas, line 20: My impression is that  "office" is a good term to describe the position of bishops or kings but an odd term for "regular" occupations (e.g. shepherd).

This isn't an issue if it reflects what the researchers have in mind, but maybe we should check that in the next meeting.

lukavdplas (approved):

backend/person/models.py
XanderVertegaal: The 'office of shepherd' sounds a bit off, indeed :P Let's ask the team on Wednesday whether they have a better solution.

XanderVertegaal:
Thanks for the clear and extensive explanation. It's good to hear your thoughts on this. I will transfer a few fixtures I've added to the tests where they are used. There is no need to 'pollute' the global `conftest.py` with these.

PR #26, opened 2024-02-09
Feature/sources

### LINKED ISSUE #21: Latin terminology
Allow adding the latin terminology used for something.

This is relevant on a lot of levels, so it probably makes sense to add this to the `Field` model, but it can also make sense for some "objects" such as letters.

### LINKED ISSUE #22: Add source information to field
Allow logging something about the provenance of the information to the Field category

lukavdplas:
This update adds a rudimentary model for describing source information.

Should close #21, close #22

This adds:
- A `Source` table to add source texts
- A `Reference` table which describes a reference in a source.

References include a GenericForeignKey field that allows them to refer to _any_ object in the database. Currently, you can add references for people, actions, and letters. We could also consider allowing specific references for individual fields.

References can include a lot of extra data to describe exactly how the source describes the object, including Latin terminology used in the text. This version is intended as a proof of concept, but we'll probably want more fields in the future.

The description of Sources is _very_ simple: just a name and a text field for details. I think it's enough for now, but we may want to expand that in the future. Ideally by reusing an existing system for bibliographies.


backend/source/models.py
```diff
@@ -0,0 +1,78 @@
...
+        max_length=200,
+        blank=True,
+        help_text="specific location of the reference in the source text",
+    )
+
+    terminology = ArrayField(
+        models.CharField(
+            max_length=200,
+        ),
+        default=[],
```
XanderVertegaal, line 61: I'm getting this warning when starting my server or applying migrations.
```
source.Reference.terminology: (fields.E010) ArrayField default should be a callable instead of an instance so that it's not shared between all field instances.
        HINT: Use a callable instead, e.g., use `list` instead of `[]`.
```

backend/source/models.py
```diff
@@ -0,0 +1,78 @@
...
+
+    mention = models.CharField(
+        max_length=32,
+        blank=True,
+        choices=[("direct", "directly mentioned"), ("implied", "implied")],
+        help_text="how is this entity presented in the text?",
+    )
+
+    def __str__(self):
+        return f"reference to {self.content_object} ({self.content_type.model}) in {self.source}"
```
XanderVertegaal, line 75: I would add `{'(' + self.location + ')' if self.location else ''}` here, but we also let the team bring it up themselves if they think it is relevant to show the specific location as well.

backend/source/models.py
```diff
@@ -0,0 +1,78 @@
...
+        on_delete=models.CASCADE,
+        help_text="the source text in which this references occurs",
+    )
+
+    # description of the reference
+
+    location = models.CharField(
+        max_length=200,
+        blank=True,
+        help_text="specific location of the reference in the source text",
```
XanderVertegaal, line 54: Most `help_text`s currently start with a capital letter. Personally, I think capitalisation looks best, but I'd be happy to use lowercase letters everywhere if you prefer. I do think we should be consistent.

XanderVertegaal (approved):
Nice job! I didn't know about GenericForeignKeys, but they are precisely what we need for the purposes of referencing. I'm suggesting a few minor changes, but those shouldn't stop you from merging.

Note to ourselves: don't forget to grant permissions to the 'Lettercraft Core Team' w.r.t. the new `Source` app after we've deployed a new version.

Issue #23, opened 2024-02-07
Add instigator role


lukavdplas:
Expand the list of action roles with "instigator"

XanderVertegaal:
Resolved with #27 .

PR #27, opened 2024-02-12
Feature/epistolary event model updates

### LINKED ISSUE #9: Model changes for EpistolaryEvent
(Note: from the presented notes, it appears that 'epistolary events' should refer to 'episodes', i.e. the collections of '(atomic) actions'.)

- The `present` box on the Role was not very useful because a single Epistolary Event could be broken up into several actions, and a person may well have been around for some but not all actions. With the new distinction between Epistolary Events + Letter Actions, this worry should be taken away.
- Add `trigger` field to indicate what triggered an epistolary event/episode: 'death of a bishop' etc. This should probably be a new model 'WorldEvent' or something like that.
- Add field `related letters`
- Add field `keywords` (Not sure whether this is a good idea, as a list of strings will be difficult to compare across events. Maybe we should use episodes for this)

~~- Add field `relevant Latin terminology`~~ (Taken up in #21)
~~- Add field `location/space`~~ (Taken up in #24)
~~- Add `instigator` field: useful for when a letter is sent, but not necessary written.~~ (Taken up in #23 )

XanderVertegaal:
Resolves #9, #23.

This PR
- adds a `Gift` model, with foreign keys to an associated 'giver' and a letter action;
- adds a `WorldEvent` model, with a foreign key to an epistolary event. In the future, the team may decide to attach world events to letter actions, but for now, epistolary events feel like a good place to start.
- adds a computed property `display_date` on the abstract `LetterCraftDate` model;
- adds `instigator` to the list of possible Roles.


backend/event/apps.py
```diff
@@ -4,4 +4,4 @@
 class EventConfig(AppConfig):
     default_auto_field = 'django.db.models.BigAutoField'
     name = 'event'
-    verbose_name = 'epistolary episodes and actions'
+    verbose_name = 'events and actions'
```
lukavdplas, line 7: I think this change is just for consistency, but in general, I think "episode/action" is clearer terminology than "event/action".

backend/event/models.py
```diff
@@ -160,3 +166,53 @@ class Role(Field, models.Model):
...
+    )
+
+    world_event = models.ForeignKey(
+        to=WorldEvent,
+        on_delete=models.CASCADE,
+        help_text="The world event that triggered an epistolary event",
+    )
+
+    def __str__(self):
+        return f"{self.world_event} triggered {self.epistolary_event}"
```
lukavdplas, line 218: Since this model doesn't contain extra metadata about the link, it looks like you don't need it; we can just use `ManyToManyField` without a `through` model.

That said, I think this should actually be an subclass of `Field` right? Then we would have extra properties.

backend/event/models.py
```diff
@@ -61,13 +60,19 @@ class LetterAction(models.Model):
...
+    )
+
     def __str__(self):
         categories = self.categories.all()
         category_names = [category.get_value_display()
                           for category in categories]
         category_desc = ", ".join(category_names)
         letters = ", ".join(letter.__str__() for letter in self.letters.all())
-        return f"{category_desc} of {letters}"
+        return f"{category_desc} of {letters} ({self.date.display_date})"
```
lukavdplas, line 75: If it's useful for the admin overview, we could also add the date to the admin list display.

backend/event/models.py
```diff
@@ -160,3 +166,53 @@ class Role(Field, models.Model):
...
+        null=False,
+        blank=True,
+        help_text="Additional notes that describe the event and its relevance to the letters.",
+    )
+
+    epistolary_events = models.ManyToManyField(
+        to=EpistolaryEvent,
+        through="EpistolaryEventTrigger",
+        related_name="triggers",
+    )
```
lukavdplas, line 194: Logically, epistolary events can also trigger world events, or epistolary events can trigger other epistolary events. Perhaps this requires a more complex relationship?

backend/event/apps.py
XanderVertegaal: My reasoning was that the term 'episodes' is otherwise not (yet) used, so it didn't make much sense to include it here. I agree that 'events' is a bit vague, so if you have any suggestions for terms that encapsulate both world and epistolary events, let me know!

backend/event/models.py
XanderVertegaal: Good idea!

backend/event/models.py
XanderVertegaal: I'm interested, but I'm not sure what you mean. Would this be a dynamic label like the date ranges you have added to the Letter admin?

backend/event/models.py
XanderVertegaal: Ah, good point. I'll make the relationship bidirectional.

backend/event/models.py
lukavdplas: Ah, I meant that you could configure the admin class of this model with something like `list_display = ['description', 'display_date']` (if `description` and `display_date` are database fields or properties with `@admin.display()` decorators). Then the admin would show the date as a second column in the overview.

XanderVertegaal:
> Ah, good point. I'll make the relationship bidirectional.

This turned out to be a bit more complicated than I thought (but good practice with working with recursive ManyToManyFields). The new structure can be summarised as follows:

- Epistolary Events can trigger World Events
- World Events can trigger Epistolary Events
- Epistolary Events can trigger other Epistolary Events
- World events can trigger other World Events

Let me know what you think!

backend/event/admin.py
```diff
@@ -71,11 +78,54 @@ class EpistolaryEventLetterActionInline(admin.StackedInline):
...
+    model = models.WorldEvent.triggered_epistolary_events.through
+    fields = ["epistolary_event", "certainty", "note"]
+    extra = 0
+    verbose_name = "Epistolary event triggered by this world event"
+    verbose_name_plural = "Epistolary events triggered by this world event"
+
+
+class WorldEventsTriggeredWorldEventsInline(admin.StackedInline):
+    model = models.WorldEvent.triggered_world_events.through
+    fk_name = "triggering_world_event"
```
XanderVertegaal, line 118: The `fk_name` is required if there are two references to the same model on the inline model.

backend/event/models.py
```diff
@@ -29,7 +29,23 @@ class EpistolaryEvent(models.Model):
...
+        through="EpistolaryEventTrigger",
+        symmetrical=False,
+        related_name="world_event_triggers",
+        help_text="World events triggered by this epistolary event",
+    )
+
+    triggered_epistolary_events = models.ManyToManyField(
+        to="self",
+        through="EpistolaryEventSelfTrigger",
+        through_fields=("triggering_epistolary_event", "triggered_epistolary_event"),
```
XanderVertegaal, line 46: Good to know (in case you didn't know this -- I didn't!):
- Without `symmetrical=False`, Django will assume that any model's ManyToMany relationship with itself is symmetrical and will add _two_ records in the database.
- `through_fields` makes sure that when you call `world_event.triggered_world_events.add(...<another-world-event>...)`, this `another-world-event` is mapped to the correct field in the through model. In other words, it marks which of the two foreign fields in the through model is the _source_ and which is the _target_.

Issue #31, opened 2024-02-23
Mock-up for data entry form


XanderVertegaal:
Data entry flow -- how would a research assistant enter new data into the database, starting from the sources.

1. Source overview => user selects/adds source
2. Entity overview (Agents / Objects / Places) => user selects/adds entity
3. Entity detail form => user enters details of agent/object/place as they occur in the source.
4. Uniquely identifiable: is the entity generic (e.g. 'a shepherd', 'a forest', 'a cask of beer') or unique identifiable ('St. Radegund', 'the monastery of Toulouse', 'the sceptre of Charlemagne')?
5. Link to existing records. (Should the research assistant do this, or will core team members do this?)

See photo below.

![image](https://github.com/CentreForDigitalHumanities/lettercraft/assets/73882506/d46c215e-0867-44d9-bc51-4edf45bdf5e9)

lukavdplas:
Current status:

#35 implements a menu to edit the data for a source, and a form to edit a depiction of an agent in a source.

I'm currently working on adding some more detail for objects and spaces. We'll try to work out actions as well in the future.

PR #35, opened 2024-02-27
data entry mockup

lukavdplas:
This is a mock-up for the data entry form (#31)

I'm making a PR because that's a good interface for review, but this should **not** be merged. This is just intended to discuss the general workflow with the researchers. It does not function.

@XanderVertegaal I think this is mostly what we already discussed, but let me know if you have thoughts or suggestions!

Included screens:
- Go to `/data-entry` to begin. This page shows an overview of your source (these would be created by an admin and assigned to research assistants, or created by the assistant themself)
- Click on the first source to see an overview of the data entered for that source. This shows agents, places, objects, and actions.
- There is also a data form for agents. Click on the "edit" button for St. Radegund to see a concept for individuals, and the "edit" buttton for the nuns to see a concept for groups.

Screenshots:

screenshot showing a list of sources

screenshot showing the menu for a source, focused on the list of agents

screenshot showing the menu for a source, focused on the list of objects

screenshot of the agent form, focused on "identification" (e.g. name)

screenshot of the agent form, focused on "description in the source"

screenshot of the agent form, focused on "connections" (asks whether this is a known historical figure)


frontend/src/app/data-entry/agent/agent.component.html
```diff
@@ -0,0 +1,245 @@
...
+                            <input class="form-check-input" type="radio" checked="true">
+                            <label class="form-check-label">
+                                Yes, this person is a historical or religious figure;
+                                they may appear in other sources
+                            </label>
+                        </div>
+                        <div class="form-check">
+                            <input class="form-check-input" type="radio">
+                            <label class="form-check-label">
+                                No, this person is an unknown individual
```
XanderVertegaal, line 237: I would lean more towards 'uniquely identifiable' and 'generic' rather than 'historical'. 'A monk' is still a historical (and religious) figure, even if we do not know their name.

frontend/src/app/data-entry/source/source.component.html
```diff
@@ -0,0 +1,254 @@
...
+                        <button class="btn btn-success">
+                            <fa-icon [icon]="icons.create"></fa-icon>
+                            Add a new object
+                        </button>
+                    </div>
+                </ng-template>
+            </div>
+        </div>
+    </div>
+    <div ngbAccordionItem>
```
XanderVertegaal, line 212: The 'Actions' are now listed in the same accordion block as the agents, objects and places, while the researchers seemed to suggest they wanted to link these in actions and events/episodes at a later stage.

We could create a second accordion, or we could just mention/ask this in the meeting, as this is a quick mock-up anyway.

XanderVertegaal (approved):
This looks great! Hopefully the team thinks it is helpful to have a more concrete visualization of what the logging flow could look like.

frontend/src/app/data-entry/source/source.component.html
lukavdplas: Yes, I wasn't sure about this. There is an "implied order" in the accordion, where actions are the final step, but we could divide it into the two stages that were proposed.

(Btw, I also used the accordion because it came straight out of the component library. It's quite possible we'll end up using something else.)

frontend/src/app/data-entry/agent/agent.component.html
lukavdplas: Good point. I'm not sure about the best terminology here, but "identifiable" is probably the clearest term we have come up with.

Issue #32, opened 2024-02-23
Split entity models into source-based and historical


XanderVertegaal:
If we decide to build upon the ideas in #31 , we can split the models for Entities (Agents/Objects/Spaces) into two separate models: one for the entity as it is described in a particular source, and one for the actual historical entity that combines all source descriptions.

PR #36, opened 2024-02-29
Feature/space models

### LINKED ISSUE #24: Location/space model
We have asked the researchers to reflect on this.

From the team's session notes:
> It might be useful to have separate categories for Space/place. Ideally, this could be done on several levels: Poitiers  Monastery of Holy Cross at Poitiers  oratory at the Monastery of Holy Cross at Poitiers. 

Question sent to the team on February 7th:
> Distinguishing between different types of locations (e.g. political or clergical areas, buildings, general descriptors like "a cave") can help to make sense of this. We can also think of a "journey" as a kind of location onto itself. I think the primary use of locations will be as something tied to actions, so think of what information you would be adding here.

lukavdplas:
This adds a new app "space" to describe spatial information. (close #24)

The core model is `SpaceDescription`, which can include anything a source text describes about where something takes place.

`Structure` is the most detailed aspect. Based on our discussion, a `Structure` describes anything from a town to a desk. Structures are connected in trees to describe how one structure contains another. This is different from the suggested implementation of `Site` and `Location` in #24, but we hope this will be clearer in data entry.

The models for `PoliticalRegion`, `EcclesiasticalRegion`, and `GeographicalRegion` models are quite minimal. Those should also include a concept of one region containing another, but I'll leave that for a later update.

Some of the ways in which spaces relate to other models and to sources will change with #32, so I did not really focus on the relationship with sources. (Descriptions should really have a single, non-nullable `source` field, but now they just reuse `Reference`.) 

**Changes for deployment:** Add the `'space'` app to settings and give staff users permission to view/edit it.


backend/space/validators.py
```diff
@@ -0,0 +1,15 @@
+from django.core.exceptions import ValidationError
+from django.db.models import Choices, IntegerChoices
+
+
+def get_label(value, options: Choices):
+    return next(label for (v, label) in options.choices if v == value)
+
+
+def validate_level_deeper_than_parent(level, parent, options: IntegerChoices):
+    if not level > parent.level:
```
XanderVertegaal, line 10: Should we allow for structures of one level to be a child of structures of the _same_ level? I tried to add the 'Poitiers nunnery' (part of a monastery, but multiple rooms, so level 3: building) with 'The Poitiers monastery' (level 3) as its parent, but that was not allowed.

I am not sure how real this is, though, so we could also simply put this as a question to Robert/Hope.

backend/space/validators.py
```diff
@@ -0,0 +1,15 @@
...
+        level_label = get_label(level, options)
+        parent_level_label = get_label(parent.level, options)
+        raise ValidationError(
+            f"cannot add a structure of level {level} ({level_label}) as a child to a structure of level {parent.level} ({parent_level_label})"
```
XanderVertegaal, line 14:
```suggestion
            f"Cannot add a structure of level {level} ({level_label}) as a child to a structure of level {parent.level} ({parent_level_label})"
```

backend/space/conftest.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+from . import models
+
+
+@pytest.fixture
+def road_structure(db):
+    return models.Structure.objects.create(
+        name="Sesame Street", level=models.Structure.LevelOptions.ROAD
```
XanderVertegaal, line 9: 😂

backend/space/models.py
```diff
@@ -0,0 +1,194 @@
...
+            [child] + child.descendants() for child in self.children.all()
+        )
+        return list(itertools.chain.from_iterable(iterate_descendants))
+
+    def clean(self):
+        if self.parent:
+            validators.validate_level_deeper_than_parent(
+                self.level, self.parent, self.LevelOptions
+            )
+
```
XanderVertegaal, line 159: Maybe you could extract the names of ancestors/descendants so they show up nicely in the Admin instead of: `Descendants: [<Structure: St. Radegund's desk>, <Structure: St. Radegund's room>]`

backend/event/models.py
```diff
@@ -84,6 +84,12 @@ class LetterAction(models.Model):
         blank=True,
     )
 
+    space_descriptions = models.ManyToManyField(
```
XanderVertegaal, line 87: When we create the actual frontend, we should add a comment here that says: "Pick the most specific description". If I select "St. Radegund's Desk", that also implies that I mean "St. Radegund's Cell" and "Monastery of Holy Cross" and "Poitiers". Else users might feel obliged to add all of these as separate space_descriptions.

backend/lettercraft/settings.py
```diff
@@ -44,6 +44,7 @@
     "letter",
     "person",
     "source",
+    "space",
```
XanderVertegaal, line 47: When we make a new deployment, we mustn't forget to give the team members access to this new app!

XanderVertegaal (changes requested):
Well done! This is a complicated model, but you have kept it neat and 'overzichtelijk'. I especially love the added 'ancestors' and 'descendants' read-only fields in the admin, and I appreciate the tests.

Apart from a few tidbits that you'll find in the comments: some important definitions of the levels 'geographical region', 'structure' etc. are now only visible in docstrings to the Python classes. They are not reflected in the admin interface when you add them to a SpaceDescription. Of course we will tell Robert/Hope in person what we mean with those terms, but it might be good to make it clear in the admin views too.

backend/event/models.py
lukavdplas: True. Though if a source says, "Radegund wrote the letter at her desk in her cell in the Monastery of the Holy Cross", I think it's appropriate to select all three.

On the other hadn, selecting only the desk would be appropriate if the source only says "Radegund wrote the letter at her desk", while the connections between Structures show that we know the location of her desk from other sources.

backend/space/validators.py
lukavdplas: Yes, I thought we might need that, but let's see what the researchers think.

We could also adjust this function so structures of _some_ levels (e.g. buildings) accept children of the same level, while others don't.

Though in the current hierarchy, I think you could add your example as:

The Poitiers Monastery (Fortification) < Nunnery (Building) < A cell (Room)

backend/space/models.py
lukavdplas: Done! They'll now display as `St. Radegund's desk, St. Radegund's room`.

XanderVertegaal (approved):

PR #40, opened 2024-03-08
idea/detail page

XanderVertegaal:
Like #35, on which this PR is based, this PR is not to be merged. It is intended as a quick-and-dirty conversation starter for our next team meeting. None of the code used here is written with maintainability/longevity/scalability in mind.

This PR features three views that exemplify how the end user would be able to browse the Lettercraft data and trace the data back to the sources. Visit [http://localhost:4200/browse](http://localhost:4200/browse) to inspect them.

The first page is an overview of available entity types. Only `agents` is clickable at the moment.

image

Second, there is a list of agents. This table currently holds only 2 agents, but this could be a massive, paginated table, with various sorting and search functionality. The number in gray marks how many sources corroborate this data point.

image

Lastly, Radegund's detail page shows you the stored information from the sources and allows you to select/deselect these.

image

### DIFF
--- a/frontend/src/app/app.module.ts
+++ b/frontend/src/app/app.module.ts
@@ -7,11 +7,17 @@ import { HomeComponent } from './home/home.component';
 import { SharedModule } from './shared/shared.module';
 import { CoreModule } from './core/core.module';
 import { DataEntryModule } from './data-entry/data-entry.module';
+import { BrowseComponent } from './browse/browse.component';
+import { AgentsComponent } from './browse/agents/agents.component';
+import { AgentDetailComponent } from './browse/agent-detail/agent-detail.component';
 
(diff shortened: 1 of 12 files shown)


lukavdplas:
Neat!

No suggestions for changes, let's show this at the next meeting.

Some thoughts:

- I would focus on the "agent detail" view, which is the most relevant to us right now.
- Re. the "browse" page: it's useful if the frontend has an "index" view where you can just see all values, which is similar to what you made here. However, as I understand, the intention is that the *primary* interface will be more a "guided tour" rather than a table of values.
- For the source references in the detail view, maybe a numbered reference ([1], [2], etc.) rather than the full title would be cleaner? Repeating the title/author creates a lot of visual clutter.

PR #44, opened 2024-03-21
data entry mockup (contd.)

lukavdplas:
This expands the mockup menu for data entry by adding forms for places and objects.

### Supported actions

In the source menu:
- Click "edit" for the first place (the abbey) to see the form for space descriptions
- Click "edit" for the first object (letter from Radegund) to see the form for letters/objects.

In the form for the abbey:
- Click the edit button (pencil icon) on the listed political region (Aquitaine)
- Click "add political region" to see the form for adding a new region in the description. In the form that opens, type in a name. Try typing in *Burgundy* to add an existing region to the description, or something else to create a new region.
- In the "structures" section, click the edit button for Poitiers to see an example of the form. This form includes a fancy visualisation of connected structures, which is is not interactive but could be.

### Screenshots

screenshot of space description form

screenshot of menu for editing a political region

screenshot of object description form


frontend/src/app/data-entry/object/object.component.html
```diff
@@ -0,0 +1,179 @@
...
+            <button ngbAccordionButton>Description in source text</button>
+        </h2>
+        <div ngbAccordionCollapse>
+            <div ngbAccordionBody>
+                <ng-template>
+                    <div class="mb-5">
+                        <h3>Sender</h3>
+                        <div class="mb-4">
+                            <label class="form-label">Who is described as the (original) sender?</label>
+                            <div class="form-check">
```
XanderVertegaal, line 78: Instead of checkboxes, this should probably become a dropdown menu in the eventual version.

(Out of scope for now, but good to think about: will we allow inline creation of agents? - If not, the user should be able to save a draft version of this object and come back to it later.)

frontend/src/app/data-entry/object/object.component.html
```diff
@@ -0,0 +1,179 @@
...
+                                <input type="checkbox" class="form-check-input" checked>
```
XanderVertegaal, line 79: Instead of checkboxes, this should probably become a dropdown menu in the eventual version.

(Out of scope for now, but good to think about: will we allow inline creation of agents? - If not, the user should be able to save a draft version of this object and come back to it later.)

frontend/src/app/data-entry/object/object.component.html
```diff
@@ -0,0 +1,179 @@
...
+                            <label class="form-label">
+                                Additional notes
+                            </label>
+                            <textarea class="form-control" rows="2"></textarea>
+                        </div>
+                    </div>
+                    <div class="mb-5">
+                        <h3>Addressee</h3>
+                        <div class="mb-4">
+                            <label class="form-label">Who is describe as the addressee?</label>
```
XanderVertegaal, line 115:
```suggestion
                            <label class="form-label">Who is described as the addressee?</label>
```

XanderVertegaal:
Apologies for the accidental push.

PR #45, opened 2024-03-24
Feature/source description split

### LINKED ISSUE #32: Split entity models into source-based and historical

XanderVertegaal:
This PR introduces a split between actual, 'historical' entities and source descriptions (closes #32 ). The implementation is as follows.

**Feel free to take over this PR and push commits on it while I am away, if it blocks your progress.**

There are two abstract models, as per your suggestion:
- `Historical` is an empty abstract class. It marks historical entities and we can use it to later implement fields that are common to all of them, if needed.
- `SourceDescription` contains fields that used to be in the `Reference` model and is inherited by source description models. As far as I can see, this class now renders the `Reference` model obsolete, so I removed it.

The idea is that each entity is represented by three model objects (taking `Agent` as an example).
 -  `AgentBase` is the base model that contains all relevant content fields. Also all ForeignKeys (for instance on `AgentName`) should refer to this model. The end user should not see this, and it has not been included in the Django admin interface.
 - `Agent` inherits from `Historical` and represents the actual 'aggregated' historical entity in our knowledge base. (We need a better term for this 😅).
 - `AgentDescription` inherits from `AgentBase` and `SourceDescription` and includes a `target` field that links it to the `Agent`. This replaces the `GenericForeignKey` that existed on the `Reference` model.

This split has been implemented for the `Agent`, `Letter`, `Gift`, `SpaceDescription` and `LetterAction` models.

### Admin interface

In the `AgentDescription` admin interface, we can easily select a target for a description:
image

In the `Agent` admin interface, connected descriptions are shown in a semi-custom tabular inline.
image

### Other issues

This PR also includes a migration that is currently missing in develop.


backend/event/models.py
```diff
@@ -148,13 +152,38 @@ def __str__(self):
...
+    An aggregate model that represents a letter action in history. This model is based on one or multiple SourceDescriptions and other sources that are not part of this database.
+    """
+
+    pass
+
+
+class LetterActionDescription(SourceDescription, LetterActionBase, models.Model):
+    """
+    A description of a letter action in a source.
+    """
```
lukavdplas, line 173: I'm not sure if there is any need for the "historical" letter action. In the case of agents, it makes sense that a description of an individual in a source will often correspond to a known historical figure, and that these historical figures are a useful "entry point" into the database. Cross-referencing individual actions is much harder, if ever possible, and less useful for making queries.

I would argue that while a `SourceDescription` is often referring to something "real" (e.g. a person, an action, a place), it's not always useful to represent both layers in the database, if they don't serve the analysis. (In the case of actions, it is also tricky that actions in texts do not necessarily represent "real" events but may serve a literary purpose.)

backend/person/models.py
```diff
@@ -150,3 +152,31 @@ class Meta:
...
+
+class Agent(Historical, AgentBase):
+    """
+    An aggregate model that represents an agent (person or group) in history. This model is based on one or multiple SourceDescriptions and other sources that are not part of this database.
+    """
+
+    pass
+
+
+class AgentDescription(SourceDescription, AgentBase):
```
lukavdplas, line 165: This is also applies to other models. In most cases, information that is represented in the `SourceDescription` does not need to be duplicated in the corresponding `Historical` model. It's more precise to display or query the agent through its descriptions.

I would say that an `Agent` may also include information of its own, but mainly things that are not really connected to sources, and more about representing the agent in the interface.

For example, the `Agent` Radegund could include a name, longer description, and an image that can be used in the interface. Then an `AgentDescription` of Radegund describes the language used in source texts to describe her gender, social status, names used in the text, etc.

backend/letter/models.py
```diff
@@ -51,7 +51,35 @@ def __str__(self):
...
+    )
+
+    def __str__(self):
+        """
+        Inherits the __str__ method from the base model and adds the source to it.
+        """
+        return f"{super().__str__()} (described in {self.source})"
+
+
+class Gift(Historical, GiftBase, models.Model):
```
lukavdplas, line 74: Same thing as for actions: I think we can start out without a `Historical` model for gifts and letters, though we can add that layer of analysis when needed.

backend/space/models.py
```diff
@@ -194,6 +195,38 @@ class LandscapeFeature(Field, models.Model):
...
+
+    pass
+
+
+class SpaceDescriptionDescription(
+    SourceDescription, SpaceDescriptionBase, models.Model
+):
+    """
+    A description of a space description in a source.
+    """
```
lukavdplas, line 218: "A description of a space description"?

This distinction is not really appropriate here; as this awkward docstring highlights, `SpaceDescription` was already a type of description.

In the case of spaces, as I understand it, source texts are so different in what they describe, and so vague, that it is a futile exercise to try to link specific places between sources. This is why we ended up with a model where a single `SpaceDescription` contains several _aspects_ that may have been described, each of which may link to a historical reference point, such as a kingdom or diocese.

So to map the `space` app to the `Historical`/`Description` division... When I wrote this originally, I imagined we would divide it like this:
- `PoliticalRegion`, `EcclesiasticalRegion`, etc., are all `Historical` models
- `SpaceDescription` is a `SourceDescription`

lukavdplas (changes requested):
Your code looks good as always, but the way the models are divided is not quite what I had in mind. I've added my thoughts here, but if you prefer, we can also discuss in person what we want to do.

XanderVertegaal:
Alright, it does seem I have misunderstood a couple of things; let's discuss the necessary refactors soon!

PR #49, opened 2024-04-11
source description split (gifts, letters)

lukavdplas:
Continuation of #48 - updates models for gifts and letters

Based on our last meeting:
- the primary focus on on `GiftDescription` and `LetterDescription`: descriptions of gifts + letters in narrative sources
- for the sake of demonstration, also added a model for `PreservedLetter`
- the models are quite minimal - you can create categories for gifts and letters, and add agents as the sender/addressee (perhaps those should also be left out?).
- existing data is removed; like in #48, there is an issue that the existing data contains both preserved and described letters.


backend/letter/models.py
```diff
@@ -94,76 +150,96 @@ def __str__(self):
...
+class PreservedLetterRole(Field, models.Model):
+    """
+    Relationship between a preserved letter and a historical figure
+    """
+
+    letter = models.ForeignKey(
+        to=PreservedLetter,
+        on_delete=models.CASCADE,
+    )
+    person = models.ForeignKey(
```
XanderVertegaal, line 239: So for the moment we are not recording what the nature of the person/agent's involvement in this letter is?

XanderVertegaal (approved):
In response to the question in the PR description: I think it's fine to leave sender/addressees in for now. I think we will need those fields anyway.

Otherwise, I just have a few clarification questions and some utterly minor suggestions for the docstrings.

PR #50, opened 2024-04-15
Refactor event models

### LINKED ISSUE #32: Split entity models into source-based and historical

lukavdplas:
Update the models for events. 

Adds `EventDescription` (i.e. an event description in a source text) -> close #32 

As agreed in our last meeting, this removes a lot of the internal structure of events, so we're not so much modelling events as just giving an overview of passages in narrative sources. I removed the modelling of causal links, as well as the idea of breaking up an event into atomic "actions".

So the result is:
- `EventDescription` describes a passage in a source text, and you can link the source's descriptions of agents/spaces/objects to it. That's basically all the internal structure it has.
- Since the `case_study` app is used for more high-level abstractions, I did add an `Episode` model there. Could still be useful for linking passages?
- Also added "key sites" and "key persons" to a case study, which is more a light suggestion. I thought this would be useful in building an interface for a case study. E.g. for the nun rebellion, the Abbey in Poitiers could be a "key site" from which you can explore the data.


lukavdplas:
Add summary field

Issue #28, opened 2024-02-16
Add common_settings file


lukavdplas:
The backend is configured using [settings.py](https://github.com/CentreForDigitalHumanities/lettercraft/blob/develop/backend/lettercraft/settings.py). In production environments, we replace that file with the one in our deployment module.

However, some settings are going to be the same across development and production, like the list of Django apps. This can lead to errors when we inevitably forget to update the deployment module after a release.

We can define shared settings in a `common_settings.py` file and import them in both the development and production settings. See https://github.com/UUDigitalHumanitieslab/I-analyzer-django/pull/51 (private repository) for an example.

jgonggrijp:
Isn't that only a partial solution, though? Often, even though the settings are not the same in development and in production, you need to change them in both places anyway. For example, you start using a setting that was not used before, and the settings needs to be different in production but it still needs to be there in order to enable some piece of functionality. Or: you need to add an item to a list in both environments, but the lists are not identical in composition.

I think any complete solutions requires looking at the diff of the settings.py at every single release. If you do that, you don't technically need a common_settings file, although I'm not opposed to adding one, either.

I will check whether the need to diff any configuration is mentioned in our documentation on how to do a release.

lukavdplas:
I mean, yes, in an ideal world, developers would always be thorough when preparing releases and there would be no human error :wink: 

In any case, a change that reduces some but not all human error is still worth it imo. Personally, I've also found that, as a solution, "I should just be more thorough from now on" is often accurate but rarely produces results.

I find a shared definition file for shared settings useful because a) you're more likely to forget about those kinds of settings, and b) those settings are _supposed_ to be the same in every instance so defining them in one place reduces the likelihood of mistakes.

This does indeed not solve anything for settings that should be different in production, but in my experience, those settings are less likely to be forgotten during deployment because developers are aware they need to be tested in production (because the situation is actually different from development).

jgonggrijp:
[...]

PR #54, opened 2024-04-25
Feature/authentication basics

lukavdplas:
This update adds basic functionality for authentication.

As discussed, the systems for authentication are based on https://github.com/UUDigitalHumanitieslab/I-analyzer. Authentication in the backend uses the dj-rest-auth en django-allauth libraries. 

### Current status

You can log in and log out in the frontend and see your login status in the top menu. You can also register a new account at `/users/registration/` (the backend endpoint); after that, you'll be able to verify your email. (In development, emails are sent to the console backend.)

The backend implements more functionality than that, since most of this could be copied from I-analyzer with little adaptation. It should provide all the necessary endpoints for:

- registering an account through the frontend
- a "forgot password" route (with a reset link sent to your email)
- changing your password (by providing your old password)
- updating your username and first name / last name 

Changing your e-mail and deleting your account isn't implemented in I-analyzer so those endpoints still have to be added.

E-mail verification is implemented, but not yet required. Making it required is trivial, but I would like to deploy this version first and make sure everything works.

The frontend is less complete, as we agreed this would be a good place to divide our labour. There is a an `AuthService` which should provide all core functionality for building components.


frontend/src/app/services/session.service.ts
```diff
@@ -0,0 +1,14 @@
...
+import { Subject } from 'rxjs';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class SessionService {
+    private static sessionExpiredSubject = new Subject<void>();
+    public expired = SessionService.sessionExpiredSubject.asObservable();
+
+    public static markExpired() {
```
XanderVertegaal, line 11: I don't see why these properties should be marked as `static`. What sets them apart from properties in, e.g., the `AuthService`? As far as I know, static properties are independent of an instance, but since services are singletons anyway, and the `Subject` here does not retain any state, the difference is quite moot (but maybe I'm missing something!).

frontend/src/app/shared/shared.module.ts
```diff
@@ -23,6 +24,8 @@ import { RouterModule } from '@angular/router';
         }),
         NgbModule,
         RouterModule,
+        FormsModule,
```
XanderVertegaal, line 27: We're not using template-driven forms, so we don't need the `FormsModule`. The project compiles fine if I remove it from `import` and `export` (also in the `shared-testing.module.ts`).

frontend/src/app/user/verify-email/verify-email.component.html
```diff
@@ -0,0 +1,27 @@
...
+    {{error.error}}
+</div>
+
+<div *ngIf="success$ | async" class="alert alert-success" role="alert">
+    <h2 class="h6 alert-heading">E-mail confirmed</h2>
+    Your email address has been verified!
+    <a *ngIf="directToLogin; else homeLink" [routerLink]="['/login']" class="alert-link">
+        Sign in
+    </a>
+    <ng-template #homeLink>
```
XanderVertegaal, line 24: I now realise the new Angular 16+ control-flow syntax would really shine here by getting rid of those silly `ng-templates`:

```html
@if (directToLogin) {
    <a [routerLink]="...">Sign in</a>
} @else {
    <a [routerLink]="...">Return to home page</a>
}
```

frontend/src/app/user/utils.ts
```diff
@@ -0,0 +1,42 @@
+import { User, UserResponse } from "./models/user";
+import _ from 'underscore';
+
+/* Transforms backend user response to User object
```
XanderVertegaal, line 4:
```suggestion
/**
/* Transforms backend user response to User object
```

XanderVertegaal: (The docstring won't show up in my IDE otherwise.)

frontend/src/environments/environment.prod.ts
```diff
@@ -4,4 +4,5 @@ export const environment = {
   production: true,
     assets: '/static/assets',
     version,
+    apiUrl: '/api'
```
XanderVertegaal, line 7: Just making sure: `environment.ts` (dev) has a trailing slash here. We probably want them to be consistent.

frontend/src/app/services/auth.service.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    public keyInfo(key: string): Observable<{ username: string, email: string }> {
+        return this.http.post<{ username: string; email: string }>(
+            this.authRoute('registration/key-info/'),
+            { key }
+        );
+    }
+
+    public showLogin(returnUrl?: string) {
+        this.currentUser$
```
XanderVertegaal, line 112: This line can probably go.

frontend/src/app/services/auth.service.ts
```diff
@@ -0,0 +1,153 @@
...
+        );
+    }
+
+    public logout(redirectToLogin: boolean = false): void {
+        this.purgeAuth();
+        this.http.post(
+            this.authRoute('logout/'),
+            {}
+        ).subscribe((data) => {
+            console.log(data);
```
XanderVertegaal, line 83: Residual console.log().

(I'll be working on this service in my PR anyway, so you may also ignore it for yours.)

backend/lettercraft/settings.py
```diff
@@ -95,6 +108,11 @@
 }
 
 
+# Email
+
+EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
```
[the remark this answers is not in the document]
lukavdplas: What do you mean with dependent? This is the module with development settings (so it also sets `DEBUG = True`). If you wanted to use a production environment with `DEBUG = False`, you should always use a separate settings module anyway.

If we want some settings to be transferred to production environments while others (like this one) are not, our most common solution is to use a common settings file (#28).

backend/user/signals.py
```diff
@@ -0,0 +1,20 @@
...
+def create_superuser_email(sender, instance, created, **kwargs):
+    """
+    Ensure newly created superuser accounts automatically have a verified email
+    """
+    if created and instance.is_superuser:
+        try:
+            EmailAddress.objects.get_or_create(
+                user=instance, email=instance.email, verified=True, primary=True
+            )
+            print(f"Automatically verified email {instance.email} for {instance}")
```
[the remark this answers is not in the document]
lukavdplas: Sure!

backend/lettercraft/settings.py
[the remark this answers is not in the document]
XanderVertegaal: Ah, nevermind then! I thought this settings.py file would also be used in production for some reason, but I now realize we actually overwrite it on the server.

frontend/src/environments/environment.prod.ts
lukavdplas: Good catch!

PR #63, opened 2024-05-04
Feature/eslint

XanderVertegaal:
Two commits:
- one adds ESLint as a dev dependency
- one fixes all existing warnings/errors

The BWI and iAnalyzer also use ESLint, so you might not object to adding it to Lettercraft as well :)

(I'd also be up for enabling Prettier, but that is more controversial/stringent, so we should discuss that beforehand.)


lukavdplas (approved):

XanderVertegaal:
Rebasing hath plunged this PR into chaos. I'll create a new branch and cherry-pick the two commits from this branch that I need.

PR #64, opened 2024-05-04
Feature/authentication frontend

XanderVertegaal:
This PR addresses a number of points listed under 'Frontend' in #38. It also introduces (NgBootstrap) toasts for app-wide notifications.

There are a few issues that could still be implemented. This PR is getting quite big, however, so I wanted to have this reviewed first before I introduce more moving parts / complexity.

## Open issues
- Return routes (these are easier to introduce once we actually have more non-auth routes than `/home`) (#68)
- The UserSettings form is very basic right now. We probably want to split it up. (#67)
- User deletion (simple DRF DestroyAPIView does not work) (#66)
- Confirmation modal for account deletion (and perhaps other destructive actions in the app)
- Route guards (not strictly necessary, but good for UX) (#68)
- A page with a privacy statement (mentioned in #38)
- A page with a terms of service statement (mentioned in #38)
- Better styling


backend/user/views.py
```diff
@@ -43,6 +43,10 @@ def post(self, request):
         except Exception as e:
             raise APIException(e)
 
+class DeleteUser(APIView):
+    def delete(self, request: HttpRequest):
```
XanderVertegaal, line 47: This does not work. I have relayed it to a separate issue.

frontend/src/app/services/auth.service.ts
XanderVertegaal, line 1: This service creates a lot of stream in very similar patterns. I played around with creating a factory function that creates a subject and four observables (for result, success, error and loading).

Let me know if you think this would be nice to have.

```ts
    private createStreams<Input, Result extends object | never = AuthAPIResult>(route: string, verb: HttpVerb): ApiStreams<Input, Result> {
        const subject = new Subject<Input>();

        const result$: Observable<Result | AuthAPIError> = subject.pipe(
			throttleTime(500),
            switchMap(input => {
                switch (verb) {
                    case 'get':
                        return this.http.get<Result>(this.authRoute(route));
                    case 'post':
                        return this.http.post<Result>(this.authRoute(route), input);
                    case 'patch':
                        return this.http.patch<Result>(this.authRoute(route), input);
                    case 'delete':
                        return this.http.delete<Result>(this.authRoute(route));
                }
            }),
            catchError(error => of<AuthAPIError>({ error: error.error })),
            share()
        );

        const success$ = result$.pipe(
            filter((result): result is Result => result ? !('error' in result) : true)
        );

        const error$ = result$.pipe(
            filter((result): result is AuthAPIError => result && 'error' in result)
        );

        const loading$ = merge(
            subject.pipe(map(() => true)),
            result$.pipe(map(() => false))
        );

        return { subject, result$, success$, error$, loading$ };
    }
```

frontend/src/app/user/login/login.component.ts
```diff
@@ -1,82 +1,73 @@
...
-        return this.usernameInput.valid && this.passwordInput.valid;
-    }
-
-    private loginSucces() {
-        this.router.navigate([this.returnUrl]);
-    }
-
-    private loginFailed(error: HttpErrorResponse) {
-        this.requestFailed = true;
+        this.authService.login$.next(this.form.getRawValue());
```
XanderVertegaal, line 71: `.getRawValue()` instead of `.value` following the docs. When using `.value`, you only take values of non-disabled controls, and since all controls can be disabled programmatically, the type is `Partial<...>`. `getRawValue()` gets all values from all controls.

frontend/src/app/user/user-settings/user-settings.component.html
```diff
@@ -0,0 +1,65 @@
...
+        </button>
+        <button class="btn btn-primary" type="button" (click)="requestPasswordReset()">
+            Request new password
+            <div *ngIf="requestResetLoading$ | async" class="spinner-border spinner-border-sm" role="status">
+                <span class="visually-hidden">Requesting password reset link...</span>
+            </div>
+        </button>
+        <button class="btn btn-danger" type="button" (click)="deleteAccount()">
+            Delete account
+            <div *ngIf="deleteUserLoading$ | async" class="spinner-border spinner-border-sm" role="status">
```
XanderVertegaal, line 60: This obviously needs a 'Are you sure you want to do this?' confirmation modal / button.

frontend/src/app/user/user-settings/user-settings.component.ts
```diff
@@ -0,0 +1,156 @@
...
+    public submit(): void {
+        this.form.markAllAsTouched();
+        updateFormValidity(this.form);
+        if (this.form.invalid) {
+            return;
+        }
+        const patchInput = this.form.getRawValue();
+
+        // If we send the username along, dj-auth-rest will assume it has changed,
+        // returning a 'username already taken' error if it hasn't.
```
XanderVertegaal, line 148: This is quite silly.

frontend/src/app/user/utils.ts
```diff
@@ -40,3 +42,124 @@ export const encodeUserData = (data: Partial<User>): Partial<UserResponse> => {
...
+        'required': 'First name is required.',
+    },
+    lastName: {
+        'required': 'Last name is required.',
+    },
+};
+
+
+/**
+ * Watches a FormControl and turns its errors into an array of string messages.
```
XanderVertegaal, line 100: This could also be implemented as a pipe.

frontend/src/main.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference types="@angular/localize" />
```
XanderVertegaal, line 1: I had to add localize again in order to use NgBootstrap's toasts. Apparently, the template NgBootstrap uses contains an `i18n-` attribute.

frontend/src/styles.scss
```diff
@@ -12,3 +12,7 @@
         padding: 0 !important;
     }
 }
+
+.always-block {
```
XanderVertegaal, line 16: Overrides `hidden: true;` from Bootstrap. It only turns visible directly after an invalid control. I'm using it for form errors.

frontend/src/app/user/register/register.component.ts
```diff
@@ -0,0 +1,100 @@
...
+            validators: [
+                Validators.required,
+                ...passwordValidators,
+            ]
+        }),
+    }, {
+        validators: identicalPasswordsValidator<keyof RegisterForm>('password1', 'password2')
+    });
+
+    public usernameErrors$ = controlErrorMessages$(this.form, 'username');
```
XanderVertegaal, line 55: Instead of creating this stream of error messages here, we could also create a pipe that extracts them from a control in the template.

That could simplify this TS file, but again, I did not want to unnecessarily introduce another layer of complexity where it's not strictly needed.

frontend/src/app/user/user-settings/user-settings.component.html
```diff
@@ -0,0 +1,65 @@
+<h2>User settings</h2>
+
+<p>Use the form below to change your name or password.</p>
+
+<form [formGroup]="form" (ngSubmit)="submit()">
```
lukavdplas, line 5: It's common that different settings connect to different models (in an applications where users have more to do) or that some things like changing your password/email have their own workflow for confirmation/verification.

So rather than using a single `<form>`, I usually to build a more modular settings menu.

For example, the user settings in LiNT are layed out like this:

screenshot of settings menu of LiNT application

My point is not about aesthetics, but the idea is to divide the page into several sections that can function as stand-alone components and use their own `<form>` (if applicable). For example, the change password option folds out into an *old password / new password / repeat new password* `<form>` that has its own submit event. This makes sense because (both here and in LiNT), changing your password connects to a different API endpoint from other settings.

Another advantage of dividing settings into several sub-components is that it's a component that can really grow out of control otherwise.

In this case, it would make sense to divide this page into the following sections:

- Username
- Personal details (i.e. first/last name)
- Email
- Password
- Delete account

frontend/src/app/user/user-settings/user-settings.component.html
```diff
@@ -0,0 +1,65 @@
...
+            <div *ngIf="updateSettingsLoading$ | async" class="spinner-border spinner-border-sm" role="status">
+                <span class="visually-hidden">Requesting password reset link...</span>
+            </div>
+        </button>
+        <button class="btn btn-primary" type="button" (click)="requestPasswordReset()">
+            Request new password
+            <div *ngIf="requestResetLoading$ | async" class="spinner-border spinner-border-sm" role="status">
+                <span class="visually-hidden">Requesting password reset link...</span>
+            </div>
+        </button>
```
lukavdplas, line 57: Since the user is already logged in (and presumably knows their password), it's preferable to use the [`/password/change` endpoint](/password/change/) here, rather than a reset link. (Though it can be good to provide a reset link as an alternative.)

frontend/src/app/user/user-settings/user-settings.component.ts
```diff
@@ -0,0 +1,156 @@
...
+        // Therefore we remove the username from the input if it has not changed.
+        if (patchInput.username === this.currentUsername) {
+            delete patchInput.username;
+        }
```
lukavdplas, line 152: This logic belongs in the `authService`, since it's about an idiosyncracy of the request. (Or, from another perspective: any component that pushes to `authService.updateSettings$` would need the same check - so this is not a quirk of this component, but of the authService.)

frontend/src/app/core/menu/user-menu/user-menu.component.ts
```diff
@@ -11,26 +10,29 @@ import _ from 'underscore';
...
-        );
-        this.user$ = this.authService.currentUser$;
-        this.showSignIn$ = this.authService.currentUser$.pipe(
-            map(_.isNull)
-        );
-    }
+    public loading$ = merge(
+        this.authService.logout$.pipe(map(() => true)),
+        this.authService.logoutResult$.pipe(map(() => false)),
+    ).pipe(startWith(false));
```
lukavdplas, line 30: It's confusing that this component has both `isLoading$` and `loading$`. Please rename to something like `authLoading$` and `logoutLoading$`.

frontend/src/app/user/login/login.component.ts
```diff
@@ -1,82 +1,73 @@
...
+    public form = new FormGroup<LoginForm>({
+        username: new FormControl<string>('baltus', {
+            nonNullable: true,
+            validators: [Validators.required]
+        }),
+        password: new FormControl<string>('knerbonius', {
+            nonNullable: true,
+            validators: [Validators.required]
+        }),
+    });
```
lukavdplas, line 30: Are `baltus` / `knerbonius` testing values you forgot to remove? The form now autofills these values, which is very weird.

frontend/src/app/core/menu/user-menu/user-menu.component.ts
XanderVertegaal: Done!

backend/user/views.py
XanderVertegaal: You're right that this needs more authentication. I've created a new issue to keep track: # 66.

frontend/src/app/services/auth.service.ts
lukavdplas: It seems like a good idea to make a general utility for this!

My suggestion would be to make this a bit more object-oriented, e.g. as a `Request` class which offers these four observables as members.

frontend/src/app/user/login/login.component.ts
XanderVertegaal: Yes, you caught this PR while it was still in draft mode that was technically not review-ready. I've obviously removed this values in a later commit (that had not been pushed yet).

(Still, they should have never made it into a commit anyway; thanks for spotting this!)

frontend/src/app/services/auth.service.ts
XanderVertegaal: Good idea!

frontend/src/app/services/auth.service.ts
```diff
@@ -1,153 +1,166 @@
...
 })
 export class AuthService {
-    private currentUserSubject$ = new BehaviorSubject<User | null | undefined>(undefined);
+    public login$ = new Subject<UserLogin>();
+    public loginResult$ = this.login$.pipe(
+        throttleTime(500),
+        switchMap(loginForm => this.http.post<AuthAPIResult>(
+            this.authRoute('login/'), loginForm
+        ).pipe(
+            catchError(error => of<AuthAPIError>({ error: error.error })),
```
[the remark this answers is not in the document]
XanderVertegaal: I agree that there are quite a lot of / too many `catchError` operators here. The `error` callback sounds good, but it requires subscribing, which is otherwise not always necessary. I'm basically using `catchError` as a quasi-`mapToError` operator, based on https://www.learnrxjs.io/learn-rxjs/operators/transformation/partition (but without using the `partition` operator itself). If you have an alternative way of obtaining a stream of errors, I'd be happy to drop `catchError`.

Once I've added the util that creates a subject + 3 streams using a class, as per your suggestion, it will hopefully look more maintainable, but I'm interested to hear what you think.

frontend/src/app/app.component.ts
```diff
@@ -1,10 +1,48 @@
...
+
+    constructor(
+        private authService: AuthService,
+        private router: Router,
+        private toastService: ToastService,
+        private destroyRef: DestroyRef,
+    ) { }
+
+    ngOnInit(): void {
+        this.authService.logoutResult$.pipe(
```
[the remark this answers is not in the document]
XanderVertegaal: I generally try to avoid subscribing in a service, but I agree with you that the `AppComponent` is not an obvious place to look for this subscription.

Can I offer you `user-menu.component.ts` as a compromise? :P (It's also the place where `logout$.next()` is called.)

frontend/src/app/user/user-settings/user-settings.component.html
XanderVertegaal: I really like this! This PR is getting big as it is, so I've created a separate issue for it: #67.

frontend/src/app/user/user-settings/user-settings.component.ts
XanderVertegaal: Good point. Done!

frontend/src/app/user/user-settings/user-settings.component.html
XanderVertegaal: I agree! Let's take this up in #67 (refactoring the UserSettings form).

XanderVertegaal: Good suggestion! I've relegated this to #67

Issue #75, opened 2024-06-11
Implement GraphQL


lukavdplas:

XanderVertegaal:
- [x] Create endpoint, set up Apollo + Graphene
 - [x] Implement types + queries for Source app
 - [x] Implement types + queries for Person app
 - [ ] Implement types + queries for Event app
 - [ ] Implement types + queries for Letter app
 - [ ] Implement types + queries for Space app
 - [ ] Implement types + queries for Person app
 - [ ] Implement types + queries for Event app
 - [ ] Implement types + queries for Letter app
 - [ ] Implement types + queries for Space app

PR #81, opened 2024-06-14
Feature/graphql scaffold

XanderVertegaal:
Part of #75 

Basic implementation of a GraphQL API using Graphene (Django) en Apollo (Angular).

Don't forget to install dependencies and run `yarn codegen` when running the application frontend.

This PR also adds a SourceType, a query `sources` and two mutations: `DeleteSource` and `UpdateOrCreateSource`, so there is something to test. Visit `localhost:8000/api/graphql` to test them out. 

Still to do (in other PRs):
- Handle CSRF correctly. I don't have much experience with this; any ideas are welcome. (#82)
- Use the correct URLs on production. For now, the endpoints in `codegen.ts` and `graphql.module.ts` are hardcoded.
- Haul over some handy utils from DIAPP (PortalDev, also uses GraphQL) that make mutations and error returning less boilerplatey.
- Add types/queries/mutations for the other objects in our data model. (I'll work on this next.)

I'll focus on writing more types and queries after this PR so we can start using them in the frontend when the views/components are there.


backend/api/graphql/schema.py
```diff
@@ -0,0 +1,18 @@
+from graphene import ObjectType, Schema
+
+from source.mutations.DeleteSourceMutation import DeleteSourceMutation
+from source.queries import SourceQueries
+
+from source.mutations.UpdateOrCreateSourceMutation import UpdateOrCreateSourceMutation
+
+
+class Query(SourceQueries, ObjectType):
```
XanderVertegaal, line 9: Not everyone is a fan of inheritance like this. We can also write all the queries (and their resolvers here), but that usually ends up being a very long list.

backend/source/queries.py
```diff
@@ -0,0 +1,8 @@
+from graphene import ObjectType
+from graphene_django import DjangoListField
+
+from source.types.SourceType import SourceType
+
+
+class SourceQueries(ObjectType):
+    sources = DjangoListField(SourceType)
```
XanderVertegaal, line 8: DjangoListField is a handy util from `graphene_django` that has a built-in resolver. By default it returns all objects in the database. When we start to add filtering/sorting, we will have to write `resolve_sources()` ourselves.

backend/source/types/SourceType.py
```diff
@@ -0,0 +1,15 @@
...
+from graphene_django import DjangoObjectType
+from source.models import Source
+
+
+class SourceType(DjangoObjectType):
+    class Meta:
+        model = Source
+
+    @classmethod
+    def get_queryset(
```
XanderVertegaal, line 12: I usually define `get_queryset()` on every type. It is used to filter the queryset, most often based on the user making the request (obtainable via `info.context.user`). It's not necessary (and in fact `return queryset.all()` is the default implementation), but it is nice to be explicit about returning _all_ objects of a particular type.

frontend/tsconfig.json
XanderVertegaal, line 1: I did not touch this file. Changes were made after running `ng add apollo-angular`.

frontend/src/app/graphql.module.ts
```diff
@@ -0,0 +1,24 @@
+import { APOLLO_OPTIONS, ApolloModule } from 'apollo-angular';
+import { HttpLink } from 'apollo-angular/http';
+import { NgModule } from '@angular/core';
+import { ApolloClientOptions, InMemoryCache } from '@apollo/client/core';
+
+const uri = 'http://localhost:8000/api/graphql'; // <-- add the URL of the GraphQL server here
```
XanderVertegaal, line 6: This URL is now hardcoded. I can move it to `environment.ts` and `environment.prod.ts`, but how does this work when we start deploying? What URL should I pick for the production backend?

I'm sure you have run into this issue before in other projects, so any advice would be helpful!

lukavdplas: In most cases, you use absolute URLs without a hostname (+ port), e.g. `'/api/graphql'`. 

If you're accessing the frontend development server on `https://localhost:4200`, this will resolve as `https://localhost:4200/api/graphql`. The development server includes a proxy configuration, which will redirect that to `https://localhost:8000/api/graphql`, i.e. the backend port.

On production, you don't have this kind of proxying, as the endpoints for the frontend client and the backend API will have the same origin (e.g. `https://myproject.org` and `https://myproject.org/api/graphql`).

The [example in the apollo-angular documentation](https://the-guild.dev/graphql/apollo-angular/docs/data/network#usage) uses a URL like this, did you try that?

backend/source/mutations/DeleteSourceMutation.py
```diff
@@ -0,0 +1,21 @@
+from graphene import Mutation, ID, Boolean, ResolveInfo, String
+from source.models import Source
+
+
+class DeleteSourceMutation(Mutation):
```
lukavdplas, line 5: The implementation of this looks fine, but deleting sources is quite destructive, so it may be better to restrict this to the admin interface.

frontend/src/app/graphql.module.ts
XanderVertegaal: Good find! I missed this in the documentation.

backend/api/apps.py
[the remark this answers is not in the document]
XanderVertegaal: It turns out `graphql` clashes with the `graphql` library. I'm now using `graphql_app` instead.

XanderVertegaal:
To address the main issue of authentication: you're right that the endpoint used to be reachable for everyone. To address this, I've added a simple Middleware (only for requests aimed at the GraphQL endpoint) that check if the user is authenticated or not. You can test this by visiting `localhost:8000/api/graphql` and querying `sources` both when you're logged out and when you're logged in. I've also added a test to makes sure this doesn't break. (Run with `python manage.py test`)

lukavdplas (approved):

PR #83, opened 2024-06-20
Feature/simplify space description

### LINKED ISSUE #59: 3 levels for spaces: Place, Region, Structure
Place (city, village)
Region (province, kingdom, Bavaria, bishopric)
Structure (castle, monastery, abbey)

lukavdplas:
close #59 

This simplifies the `SpaceDescription` model.

- A space description no contains three features; `regions`, `settlements`, and `structures`. Landscape features are removed.
- The models for political, ecclesiastical, and geographical regions are merged into a single `Region` model.
- Settlement-level structures are moved into a new model `Settlement`. Structures can belong to a settlement, but the more complex relation of structures belonging to higher-level structures is removed. (So everything within one settlement will be a flat list.)


backend/space/conftest.py
```diff
@@ -4,14 +4,25 @@
 
 
 @pytest.fixture
-def road_structure(db):
-    return models.Structure.objects.create(
-        name="Sesame Street", level=models.Structure.LevelOptions.ROAD
+def region(db):
+    return models.Region.objects.create(
+        name="Muppetopia",
```
XanderVertegaal, line 9: 😂😂

PR #86, opened 2024-06-22
Feature/person types

### LINKED ISSUE #78: Implement Agent form

XanderVertegaal:
Another part of #75 

This PR adds GraphQL types (and 2 queries) for models in Core and Person.

It also adds two queries for AgentDescriptions that may be used when we start building the forms. More queries can easily be added as we need them.

I recommend that you test this by visiting the _Graphiql_ endpoint, inspecting the schema and making test queries.

image


backend/core/types/DateFieldType.py
```diff
@@ -0,0 +1,25 @@
...
+from core.models import LettercraftDate
+from core.types.AbstractDjangoObjectType import AbstractDjangoObjectType
+
+
+class LettercraftDateType(AbstractDjangoObjectType):
+    """
+    Type for models that extend the LettercraftDate model.
+    Should not be queried directly, but should be extended by other types.
+    """
+    display_date = String(required=True)
```
XanderVertegaal, line 12: A computed `@property` field that Graphene does not know how to resolve, so we need to write our own resolver function.

backend/person/models.py
```diff
@@ -57,10 +57,12 @@ class PersonReference(Field, models.Model):
     person = models.ForeignKey(
         to=HistoricalPerson,
         on_delete=models.CASCADE,
+        related_name="person_references",
```
lukavdplas, line 60: For the HistoricalPerson model, `description_references` may be more appropriate?

lukavdplas (approved):

backend/person/models.py
XanderVertegaal: Agreed!

PR #98, opened 2024-07-25
Feature/icon library

lukavdplas:
Some icon things:

- We discussed using empty/filled icons to distinguish between known/unknown entities (implemented in #97). This is an issue with the fontawesome icon library, though: the necessary icon variants are not always free. This update switches to  [bootstrap icons](https://icons.getbootstrap.com/) for that reason.
- Added an `IconComponent` to make icons in templates more convenient.
- Added an `icons.ts` definition file, so we can have a single source of truth for what symbols are used for what concepts. (See [icons.ts in I-analyzer](https://github.com/UUDigitalHumanitieslab/I-analyzer/blob/develop/frontend/src/app/shared/icons.ts) for a more mature example.)


XanderVertegaal (approved):
Neat, looks good!

I have two small suggestions, but nothing to prevent you from merging.

frontend/src/app/shared/icon/icon.component.ts
```diff
@@ -0,0 +1,33 @@
...
+ * </button>
+ * ```
+ */
+@Component({
+    selector: 'lc-icon',
+    templateUrl: './icon.component.html',
+})
+export class IconComponent {
+    @Input() icon: string | undefined;
+}
```
XanderVertegaal, line 33: For documentation purposes and compile-time checks:

```suggestion
    @Input({ required: true }) icon: string | undefined;
```

(I would personally also use `icon!: string;` and get rid of the `undefined` and the `*ngIf` check in the template, since we don't expect this component to be used without an icon, but I know opinions differ on the use of `!`.)

PR #100, opened 2024-07-29
Feature/source list

### LINKED ISSUE #76: Implement Sources overview

XanderVertegaal:
(PR 100! 🥳)

Closes #76 

This PR contains the following:
- a DataEntryModule for the data entry forms we'll be adding this week;
- a SourcesList component that queries Sources from the database and displays them;
- a simple BreadCrumb component.

It also expands the SourceType a little bit, so that we are now also able to query `Source.written_date` and `Source.contents_date`.

In the future, we might want the BreadCrumb component to be a bit 'smarter' so it parses the URL itself or otherwise requires less manual configuration. This is best left for another issue, though.

image


frontend/src/app/core/menu/menu.component.html
```diff
@@ -19,6 +19,11 @@
                         Home
                     </a>
                 </li>
+                <li class="nav-item">
+                    <a [routerLink]="['/data-entry']" routerLinkActive="active" class="nav-link">
```
lukavdplas, line 23: Can you add an `*ngIf=isAuthenticated` here?

frontend/src/app/data-entry/sources/sources.component.html
```diff
@@ -0,0 +1,26 @@
...
+        <p class="mb-1">
+            <ng-container *ngIf="source.medievalAuthor || source.medievalTitle">
+                {{ source.medievalAuthor }}<ng-container *ngIf="source.medievalAuthor && source.medievalTitle">, </ng-container><em>{{ source.medievalTitle }}</em>.
+            </ng-container>
+            <ng-container *ngIf="source.editionAuthor || source.editionTitle">
+                (Ed. {{ source.editionAuthor }}<ng-container *ngIf="source.editionAuthor && source.editionTitle">, </ng-container><em>{{ source.editionTitle }}</em>.)
+            </ng-container>
+        </p>
+    </a>
+</div>
```
lukavdplas, line 26: Two semantics issues:

- The header level in line 12 is incorrect; this would be `<h2>`. Though in this case, I think it may be more appropriate to code this section as a list (with `<ul>`/`<li>`) than with section headers.
- It's generally recommended that an anchor (`<a>`) does not contain too much text; this makes navigation by links harder. Some accessibility checkers will also complain about putting a header inside the link (you normally do it the other way around). In this case, you could  just put the `<a>` tag around the source name in line 12, and use a [stretched link](https://getbootstrap.com/docs/5.3/helpers/stretched-link/) to increase the clickable area.

So I would suggest something like this (please double check the layout, though):

```suggestion
<ul class="list-group">
    <li *ngFor="let source of sources$ | async; trackBy: identify"
       class="list-group-item list-group-item-action">
        <div class="d-flex align-items-center">
            <a [routerLink]="['source', source.id]" class="mb-1 stretched-link">{{ source.name }}</a>
            <small class="text-body-secondary ms-2">
                ({source.numOfEpisodes, plural, =0 {No episodes} =1 {1 episode} other {{{ source.numOfEpisodes}} episodes}})
            </small>
        </div>
        <p class="mb-1">
            <ng-container *ngIf="source.medievalAuthor || source.medievalTitle">
                {{ source.medievalAuthor }}<ng-container *ngIf="source.medievalAuthor && source.medievalTitle">, </ng-container><em>{{ source.medievalTitle }}</em>.
            </ng-container>
            <ng-container *ngIf="source.editionAuthor || source.editionTitle">
                (Ed. {{ source.editionAuthor }}<ng-container *ngIf="source.editionAuthor && source.editionTitle">, </ng-container><em>{{ source.editionTitle }}</em>.)
            </ng-container>
        </p>
    </li>
</ul>
```

frontend/src/app/shared/breadcrumb/breadcrumb.component.html
```diff
@@ -0,0 +1,12 @@
+<nav>
```
lukavdplas, line 1: Since this is not the only `<nav>` element on the page, [navigation elements should be labelled](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#labeling_section_content). I think you'll have to add a label in the navbar as well.

lukavdplas (changes requested):
Neat!

The API part all looks fine to me, but I added some suggestions for the HTML templates.

frontend/src/app/data-entry/sources/sources.component.html
XanderVertegaal: I swear I will actually remember to use the correct header hierarchy one of these days :P Both the recommendation of using shorter texts in anchor tags and Bootstrap's stretched link were new to me. Good to know! 👍

frontend/src/app/shared/breadcrumb/breadcrumb.component.html
XanderVertegaal: The main navbar has already been labelled, so no changes were required in that component. I have added an `aria-label` to this breadcrumb `nav` element.

frontend/src/app/data-entry/sources/sources.component.html
XanderVertegaal: I checked the layout and it still looks good :)

lukavdplas (approved):

PR #102, opened 2024-08-01
Feature/source detail

XanderVertegaal:
Adds a source detail view and makes sure episodes are queryable on SourceType.

The new/edit form and the associated mutations will be added in other PRs.


backend/person/types/AgentDescriptionType.py
```diff
@@ -40,4 +40,4 @@ def resolve_describes(
     def resolve_person_references(
         parent: AgentDescription, info: ResolveInfo
     ) -> QuerySet[PersonReference]:
-        return parent.person_references.all()  # type: ignore
+        return PersonReference.objects.filter(person=parent)
```
XanderVertegaal, line 43: This removes the `# type: ignore`. (No functional change.)

frontend/src/app/data-entry/source/source.component.scss
```diff
@@ -0,0 +1,4 @@
+.entity-item:not(:last-child)::after {
+    // Character \00A0 is a non-breaking space.
+    content: ",\00A0";
+}
```
lukavdplas, line 4: Oh neat, this is much cleaner than my solution! :)

Perhaps we could generalise this into a more general rule that we can reuse in the application? E.g. `.inline-list > .inline-list-item`

frontend/src/app/data-entry/source/source.component.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { SourceComponent } from "./source.component";
+
+describe("SourceComponent", () => {
+    let component: SourceComponent;
+    let fixture: ComponentFixture<SourceComponent>;
+
+    beforeEach(() => {
+        TestBed.configureTestingModule({
+            declarations: [SourceComponent],
```
lukavdplas, line 11: I think this needs to import the `SharedTestingModule`

lukavdplas (approved):
Neat!

I added some minor suggestions, but none of them are hard requirements.

frontend/src/app/data-entry/source/source.component.scss
XanderVertegaal: Good idea!

frontend/src/app/data-entry/source/source.component.html
```diff
@@ -0,0 +1,129 @@
...
+                        <span>chapter {{ episode.chapter }}</span>
+                    </small>
+                </div>
+                <lc-action-button-group
+                    [editLink]="['episode', episode.id]"
+                    (delete)="deleteEpisode(episode.id)"
+                    [showButtonText]="false"
+                />
+            </div>
+        </div>
```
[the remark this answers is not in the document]
XanderVertegaal: Good tip! Done!

PR #104, opened 2024-08-02
Feature/agent location api

lukavdplas:
Two minor backend changes:

- Agent-location relations are now one-to-one (i.e. agents cannot have multiple locations)
- Adds the agent's location to the AgentDescription query in the GraphQL API


XanderVertegaal (approved):
Nice! The only thing that I could see as a potential cause for confusion is that we now have a nested `location` field, one for the `AgentDescriptionLocation` through model and one for the `SpaceDescription` field on that model.

image

I'm struggling to find a good naming alternative to disambiguate the two, but if you're not bothered by it, then feel free to merge.

PR #105, opened 2024-08-05
Bugfix/gh actions

lukavdplas:
This gets the unit test github action working again.

I spent some time trying to get code generation to work, but ended up giving up and going for the "dumb" solution to include generated types in version control.

I did clean up the github action script a bit. I removed some unnecessary commands, updated the runner + modules, added caching for dependencies. Also, the frontend/backend tests are now run in a parallel :rocket:


XanderVertegaal:
Well done, thanks for looking at this!

Having to check in the generated files is not perfect for future PRs, but it is indeed the easiest solution for now (easier than starting a temporary backend server within GitHub actions), so I'm on board with this 👍 

Do we still need the `yarn codegen-once` command (and the corresponding documentation) with this solution?

package.json
```diff
@@ -33,7 +33,9 @@
         "static-p": "yarn build && yarn clean-static && yarn collectstatic-p",
         "watch-front-p": "yarn front yarn watch",
         "start-back-p": "cd backend && python manage.py runserver --settings production --insecure --pythonpath ..",
-        "start-p": "yarn static-p && yarn watch-front-p & yarn start-back-p"
+        "start-p": "yarn static-p && yarn watch-front-p & yarn start-back-p",
+        "stop-back": "kill $(lsof -t -i:8000)",
+        "codegen-once": "yarn start-back & yarn front yarn graphql-codegen && yarn stop-back"
```
XanderVertegaal, line 38: On my system, running `yarn codegen-once` simply starts the backend server without (visibly) running codegen or quitting. Maybe this is because `lsof` does not work on Windows. I'm happy to leave it as is, but the documentation should probably mention that Windows users are out of luck in this respect :P

lukavdplas:
> Do we still need the yarn codegen-once command (and the corresponding documentation) with this solution?

Not really necessary, but it can be useful.

This comment made me realise that this branch is an out-of-date version, not the newer version with more useful commit names + documentation. I'll close this PR and merge the other version. (Which contains the same code.)

PR #106, opened 2024-08-06
Form placeholders

lukavdplas:
This creates some boilerplate pages for agents, letters, gifts and locations. Each page is just a header and a message that the form isn't there yet.

This is mostly so we can have working links while we're still working on these forms.


frontend/src/app/data-entry/agent-form/agent-form.component.html
```diff
@@ -0,0 +1,21 @@
+<ng-container *ngIf="data$ | async as data">
+    <lc-breadcrumb [breadcrumbs]="getBreadcrumbs(data)"></lc-breadcrumb>
```
XanderVertegaal, line 2: In my personal opinion, self-closing tags look just a bit neater if there are no slots involved.

```suggestion
    <lc-breadcrumb [breadcrumbs]="getBreadcrumbs(data)" />
```

frontend/src/app/data-entry/agent-form/agent-form.component.ts
```diff
@@ -0,0 +1,51 @@
...
+        this.id$ = this.route.params.pipe(
+            map(params => params['id']),
+        );
+        this.data$ = this.id$.pipe(
+            switchMap(id => this.agentQuery.watch({ id }).valueChanges),
+            map(result => result.data),
+        );
+    }
+
+    getBreadcrumbs(data: DataEntryAgentQuery): Breadcrumb[] {
```
XanderVertegaal, line 29: Part of me is a little bit worried that in some circumstances (heavy queries, bad internet connection) the query data may arrive after the DOM has been built and the breadcrumbs won't get updated. What would you say of using an observable (with `async` in the template) or computed signal (with `toSignal(...)`) for this instead to make sure they are fully reactive?

```typescript
breadcrumbs$: Observable<BreadCrumb[]> = this.data$.pipe(
    map(data => ({ ... }))
)
```

or:

frontend/src/app/data-entry/agent-form/agent-form.component.ts
```diff
@@ -0,0 +1,51 @@
...
+        if (data.agentDescription) {
+            return [
+                { link: '/', label: 'Lettercraft' },
```
XanderVertegaal, line 32: Part of me is a little bit worried that in some circumstances (heavy queries, bad internet connection) the query data may arrive after the DOM has been built and the breadcrumbs won't get updated. What would you say of using an observable (with `async` in the template) or computed signal (with `toSignal(...)`) for this instead to make sure they are fully reactive?

```typescript
breadcrumbs$: Observable<BreadCrumb[]> = this.data$.pipe(
    map(data => ({ ... }))
)
```

XanderVertegaal (approved):
Nice work! I have two style suggestions, but nothing major otherwise. :)

Issue #94, opened 2024-07-17
Log user contributions


lukavdplas:
The database should log which users contributed to a record.

This is primarily so researchers / research assistants can show their contributions to the database. For instance, they could link to their own contributions on a CV. However, this is also useful information for project management.

For these purposes, it's not necessary to keep a detailed changelog; a list of contributors to records would be fine.

Suggested implementation:
- For key models (AgentDescription, Episode, etc.), add a `contributors` field, which is a many-to-many relationship with users.
- On mutation requests, add the request user to the contributors (if they aren't in there already).

PR #107, opened 2024-08-06
Feature/contributions in db

lukavdplas:
This is a partial solution to #94 

- Adds `contributors` fields to the core data models
- You can edit contributors in the admin interface
- You can query contributors in the graphql api
- You can query users in the graphql api


backend/user/queries.py
```diff
@@ -0,0 +1,29 @@
+from graphene import Field, List, ResolveInfo
+from graphene_django import DjangoObjectType
```
XanderVertegaal, line 2: Unused imports.

backend/letter/admin.py
```diff
@@ -52,7 +52,8 @@ class PreservedLetterRoleAdmin(admin.StackedInline):
 @admin.register(models.PreservedLetter)
 class PreservedLetterAdmin(admin.ModelAdmin):
     list_display = ["name", "description"]
-    fields = ["name", "description"]
+    fields = ["name", "description", "contributors"]
```
XanderVertegaal, line 55: Instead of these fields, you can also use the existing core fieldsets (just as you did for `HistoricalPerson`). Not a hard requirement, just pointing it out for consistency's sake.

```py
    fieldsets = [
        core_admin.named_fieldset,
        core_admin.contributions_fieldset,
    ]
```

XanderVertegaal (approved):
Nice! Good idea to add contributors as a field to the EntityDescription model :)

Apart from the field definition on the EntityDescriptionType, I have no noteworthy comments.

PR #108, opened 2024-08-07
Feature/agent form

lukavdplas:
This adds a basic form for agent descriptions.

This creates a form for an agent's name, description, designators, gender, and location (for groups). The form loads data from the backend, but it doesn't yet have a method to make updates.

screenshot of agent form


frontend/src/app/data-entry/agent-form/agent-description-form/agent-description-form.component.html
```diff
@@ -0,0 +1,98 @@
...
+    <div class="mb-4">
+        <label class="form-label" for="input-gender-notes">
+            Additional notes
+        </label>
+        <textarea class="form-control" id="input-gender-notes"
+            [formControl]="form.controls.gender.controls.note">
+        </textarea>
+    </div>
+
+    <ng-container *ngIf="isGroup$ | async">
```
XanderVertegaal, line 43: Currently, the location form does not appear when I select `isGroup`. It only determines whether to hide/show the form on initial load.

frontend/src/app/data-entry/agent-form/agent-description-form/agent-description-form.component.ts
```diff
@@ -0,0 +1,112 @@
...
+            map(result => result.agentDescription?.isGroup || false),
+        );
+        this.locations$ = this.data$.pipe(
+            map(data => data.agentDescription?.source.id),
+            filter(_.negate(_.isUndefined)),
+            switchMap(id => this.locationsQuery.watch({ id }).valueChanges),
+            map(result => result.data),
+            shareReplay(1),
+        );
+        this.data$.subscribe(this.updateFormData.bind(this));
```
XanderVertegaal, line 82: See my comment in `agent-identification-form.component.ts`. Your query observables are still active after component unmounting.

frontend/src/app/data-entry/agent-form/agent-form.component.html
```diff
@@ -14,8 +14,32 @@ <h1 class="mb-4">
...
+
+        <h2>Historical context</h2>
+
+        <p><i>Coming soon!</i></p>
     </ng-container>
+
+
+    <ng-template #noAgent>
+        <div class="alert alert-danger" role="alert">
+            Cannot find agent with ID {{id$ | async}}.
```
XanderVertegaal, line 35: Nice detail! I'll be sure to steal this. 👨‍💻

XanderVertegaal (approved):
Very neat, well done! There is just one memory leak issue I'd like you to look at, but otherwise, you're good to go!

frontend/src/app/data-entry/agent-form/agent-description-form/agent-description-form.component.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    ngOnChanges(changes: SimpleChanges): void {
+        if (changes['id'] && this.id) {
+            this.id$.next(this.id);
+        }
+    }
```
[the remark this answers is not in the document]
XanderVertegaal: One more thing I just realised: instead of passing the ID around via inputs, you could have every sub-form just take it from the route again as you do it in the main form. That gets you rid of this `ngOnChanges` and the subject.

frontend/src/app/data-entry/agent-form/agent-description-form/agent-description-form.component.html
lukavdplas: True, there is a similar issue where the agent's name in the title / breadcrumb does not update when you change it in the form.

The reason is that the data is fetched from the API but the form isn't sending data yet. I think the best way to resolve this is by controlling when Apollo refreshes query results, right?

PR #111, opened 2024-08-10
Feature/update episode form

XanderVertegaal:
This PR contains basic scaffolding for the EpisodeForm, split up into various subforms. Each subform performs a small query for the data that it needs, but they all share a single general update mutation that is located at the main form (EpisodeFormComponent).

Data is loaded from the database and updated live as the user modifies the fields (with a 300 ms debounce time) for most fields. 

In addition, this PR contains a several generally useful parts (taken from the DIAPP, mostly).
- `LettercraftMutation`: this is an extension of Graphene's standard `Mutation` class. If you provide `django_model`, you can use the `get_or_create_object` and `mutate_object`, greatly reducing the boilerplate needed to update/create database objects based on mutation input. Check out the `UpdateCreateEpisodeMutation` for a demonstration.
- `LettercraftErrorType`: a simple wrapper around Graphene-Django's `ErrorType` that accepts arguments in its constructor. This is simply here to reduce undue type warnings.

Not yet implemented:
- The LabelSelectComponent and the various M2M-fields (agents, objects, locations) will follow in other PRs.
- The subforms now duplicate a lot of code, but for reasons of time and to keep things simple I have decided not to refactor just yet.
- Creating / deleting


lukavdplas:
> The only required field on an Episode is name. Currently it is marked as non-required in the mutation input, but marking it as obligatory would require each subform to submit it with every mutation. We should think about what we want to do here.

My intuition would be: if forms need to request information only to send it back unchanged, you lose some of the flexibility that you would otherwise have with GraphQL.

So I would say that mutations should only include an ID (if applicable) and the data that you actually want to mutate. So you could send mutation data like:

```
{ id: "1", summary: "Interesting things happen." }
```

Which suggests that you're only editing the `summary` property. That `name` is required on the model, would mean that this data would be invalid:

```
{ id: "1", name: null, summary: "Interesting things happen."  }
```

However, I found that this can be a bit tricky on the backend side because python doesn't have a `null`/`undefined` distinction.

XanderVertegaal:
As discussed the `UpdateCreate` mutation has been converted to an `Update` mutation. A separate `Create` mutation will be added in a new PR, along with a separate mini-form that has all the fields needed to create a new Episode.

This solves the problem of required fields on the model: they are not required for the updating mutation, but they should be for the creating mutation.

PR #114, opened 2024-08-20
Feature/page titles

lukavdplas:
Add appropriate titles for pages in the frontend.

### DIFF
--- a/frontend/src/app/titles.ts
+++ b/frontend/src/app/titles.ts
@@ -43,9 +43,10 @@ export const sourceFormTitleResolver = queryTitleResolver(
     sourceTitleQuery, sourceFormTitle);
 
 
-type AgentTitleQueryData = {
-    agentDescription?: { name: string, source: { name: string } }
-};
+type EntityDescriptionTitleQueryData<D extends string> =
+    Record<D, { name: string, source: { name: string } }>;
+
+type AgentTitleQueryData = EntityDescriptionTitleQueryData<'agentDescription'>;
 
 const agentTitleQuery = (params: Params) => gql<AgentTitleQueryData, unknown>(`
     query AgentTitleQuery {
@@ -66,9 +67,7 @@ const agentFormTitle = (data: AgentTitleQueryData) => {
 
 export const agentFormTitleResolver = queryTitleResolver(agentTitleQuery, agentFormTitle);
 
-type LetterTitleQueryData = {
-    letterDescription?: { name: string, source: { name: string } };
-};
+type LetterTitleQueryData = EntityDescriptionTitleQueryData<'letterDescription'>;
 
 const letterTitleQuery = (params: Params) => gql<LetterTitleQueryData, unknown>(`
     query LetterTitleQuery {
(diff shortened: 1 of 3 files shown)


frontend/src/app/titles.ts
```diff
@@ -0,0 +1,121 @@
...
+            map(titleFromData),
+            map(pageTitle),
+        );
+    }
+}
+
+type SourceTitleQueryData = { source?: { name: string } }
+
+const sourceTitleQuery = (params: Params) => gql<SourceTitleQueryData, unknown>(`
+    query SourceTitleQuery {
```
XanderVertegaal, line 34: I think it's better here to use `.query()` rather than `.watchQuery().valueChanges()`. You only eed to perform this query once, and it auto-completes once a result is in. This way there is less risk of a memory leak.

frontend/src/app/titles.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export const giftFormTitleResolver = queryTitleResolver(
+    giftTitleQuery, entityDescriptionFormTitle);
+
+type SpaceTitleQueryData = EntityDescriptionTitleQueryData<'spaceDescription'>;
+
+const spaceTitleQuery = (params: Params) => gql<SpaceTitleQueryData, unknown>(`
+    query SpaceTitleQuery {
+        spaceDescription(id: "${params['id']}") {
+            id
```
XanderVertegaal, line 113: I needed to look up what this "${...}" syntax was doing here, but I understand now. Nice :)

frontend/src/app/titles.ts
```diff
@@ -0,0 +1,121 @@
...
+            name
+            source { id, name }
```
XanderVertegaal, line 115: I'd prefer it if each property were on a new line. It's not required by GraphQL in any way, but it's common practice.

frontend/src/app/titles.ts
XanderVertegaal, line 34: Instead of your own defined types (`SourceTitleQuery` et al.) I would import the generated one (i.c. `SourceTitleQueryQuery`). You can also import the generated type for the variables if you want to replace the `unknown`.

frontend/src/app/titles.ts
```diff
@@ -0,0 +1,121 @@
...
+        source(id: "${params['id']}") {
+            id
+            name
+        }
+}`);
+
+const sourceFormTitle = (data: SourceTitleQueryData) => `Edit ${data.source?.name}`;
+
+export const sourceFormTitleResolver = queryTitleResolver(
+    sourceTitleQuery, sourceFormTitle);
```
XanderVertegaal, line 44: How about this (suggested by my formatter)?

```suggestion
export const sourceFormTitleResolver = queryTitleResolver(
    sourceTitleQuery, sourceFormTitle
);
```

XanderVertegaal (approved):
Very neat and DRY! I love it :)

I have a few type/formatting related suggestions, nothing functional, so go ahead and merge when you think it's ready!

PR #115, opened 2024-08-21
filter admin inline fields

lukavdplas:
Small update to filter the querysets of some fields in the admin. E.g. when you select agents in an episode, you'll only see agents from the same source text. (As other agents are not valid choices.)

I was looking into this more out of curiosity than because it's high priority at the moment, but it does make the admin site more user-friendly.

### DIFF
--- a/backend/core/admin.py
+++ b/backend/core/admin.py
@@ -1,4 +1,9 @@
 from django.contrib import admin
+from django.db.models.fields.related import RelatedField
+from django.db.models.query import QuerySet
+from django.http import HttpRequest
+import re
+
 
 named_fieldset = (
     "Name and description",
@@ -37,3 +42,13 @@ class EntityDescriptionAdmin(admin.ModelAdmin):
         contributions_fieldset,
         description_source_fieldset,
     ]
+
+
+def get_queryset_matching_parent_source(
+    admin: admin.StackedInline | admin.TabularInline,
+    db_field: RelatedField,
+    request: HttpRequest,
+) -> QuerySet:
+    parent_id = re.search(f"\d+", request.path).group(0)
+    parent = admin.parent_model.objects.get(id=parent_id)
+    return db_field.remote_field.model.objects.filter(source=parent.source)
--- a/backend/core/admin.py
+++ b/backend/core/admin.py
@@ -49,6 +49,10 @@ def get_queryset_matching_parent_source(
     db_field: RelatedField,
     request: HttpRequest,
 ) -> QuerySet:
-    parent_id = re.search(f"\d+", request.path).group(0)
-    parent = admin.parent_model.objects.get(id=parent_id)
-    return db_field.remote_field.model.objects.filter(source=parent.source)
+    id_match = re.search(r"\d+", request.path)
+    if id_match:
+        parent_id = id_match.group(0)
+        parent = admin.parent_model.objects.get(id=parent_id)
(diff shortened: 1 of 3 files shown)


backend/core/admin.py
```diff
@@ -37,3 +42,13 @@ class EntityDescriptionAdmin(admin.ModelAdmin):
...
         description_source_fieldset,
     ]
+
+
+def get_queryset_matching_parent_source(
+    admin: admin.StackedInline | admin.TabularInline,
+    db_field: RelatedField,
+    request: HttpRequest,
+) -> QuerySet:
+    parent_id = re.search(f"\d+", request.path).group(0)
```
XanderVertegaal, line 52: I'm getting an AttributeError when click `Add Agent Description` in Django Admin, because `re.search()` can't find a match.

Suggestion:
```python
    re_result = re.search(r"\d+", request.path)
    if re_result is None:
        return db_field.remote_field.model.objects.none() # or .all() depending on what you want the default to be.
    parent_id = re_result.group(0)
```

Also note that instead of an `f`-string, you probably want to use an `r` string in your regex. (Pylance is warning me about this: `Unsupported escape sequence in string literal`).

backend/person/admin.py
```diff
@@ -42,6 +47,14 @@ class AgentDescriptionLocationAdmin(admin.StackedInline):
     fields = ["location"] + core_admin.description_field_fields
     extra = 0
 
+    def get_field_queryset(
+        self, db, db_field: RelatedField, request: HttpRequest
+    ) -> QuerySet:
```
XanderVertegaal, line 52:
```suggestion
    ) -> QuerySet | None:
```

backend/person/admin.py
```diff
@@ -42,6 +47,14 @@ class AgentDescriptionLocationAdmin(admin.StackedInline):
...
+        if db_field.name == "location":
+            core_admin.get_queryset_matching_parent_source(self, db_field, request)
```
XanderVertegaal, line 54: Did you forget a `return` here?

backend/event/admin.py
```diff
@@ -1,3 +1,4 @@
+from typing import Any
```
XanderVertegaal, line 1: Unused import.

backend/person/admin.py
```diff
@@ -42,6 +47,14 @@ class AgentDescriptionLocationAdmin(admin.StackedInline):
     fields = ["location"] + core_admin.description_field_fields
     extra = 0
 
+    def get_field_queryset(
+        self, db, db_field: RelatedField, request: HttpRequest
+    ) -> QuerySet:
+        if db_field.name == "location":
```
XanderVertegaal, line 53: The type hints here hide the fact that request is typed as `HttpRequest | None`. I have no idea when this would be `None`, but best to handle this case anyway.

backend/person/admin.py
```diff
@@ -1,4 +1,9 @@
+from typing import Any
 from django.contrib import admin
+from django.db.models.fields.related import RelatedField
+from django.db.models.query import QuerySet
+from django.http import HttpRequest
+import re
```
XanderVertegaal, line 6: 2x unused import (`Any` and `re`).

XanderVertegaal (changes requested):
If I avoid the AttributeError, it works well! Have a look at that one and some typing issues, and feel free to merge afterwards.

PR #117, opened 2024-08-23
Feature/contributions

XanderVertegaal:
This PR:
- Adds users to `contributors` upon updating an episode.
- Adds API tests to make sure these additions are handled correctly.
- Adds a reusable `contributors` component to `SharedModule` in the frontend (+ tests)

image

### DIFF
--- a/backend/core/tests/test_contributors.py
+++ b/backend/core/tests/test_contributors.py
@@ -1,0 +1,239 @@
+import json
+from graphene_django.utils.testing import GraphQLTestCase
+
+from event.models import Episode
+from source.models import Source
+from user.models import User
+
+
+class ContributorTestCase(GraphQLTestCase):
+    GRAPHQL_URL = "/api/graphql"
+
+    # Any mutation involving a model that inherits EntityDescription will do.
+    # For these tests, we'll use the Episode model.
+
+    UPDATE_EPISODE_MUTATION = """
+        mutation UPDATE_EPISODE($input: UpdateEpisodeMutationInput!) {
+            updateEpisode(input: $input) {
+                ok
+                errors {
+                    field
+                    messages
+                }
+            }
+        }
+    """
+
+    @classmethod
+    def setUpTestData(cls) -> None:
+        source = Source.objects.create(name="A New Hope")
+        cls.episode = Episode.objects.create(
+            name="Death Star destroyed", source_id=source.pk
+        )
+        cls.user_1 = User.objects.create(username="Yoda")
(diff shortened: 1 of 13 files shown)


backend/core/tests/test_contributors.py
lukavdplas, line 1: These tests are on the long side for unit tests. This isn't really major issue, just a comment. But I would say that ideally, unit tests are kept short and simple.

One of the reasons why the tests are long is because you repeatedly assert the same things between tests. For instance, `test_single_contributor_one_contribution` makes a single request and checks the output and result. Then `test_one_contributor_multiple_contributions` starts by making the same request, and repeats many of the same assertions on the output and results. You can skip all of these assertions, because there is already a test that checks the output of that request. You can move straight to adding a second contribution, which is what the second test is about.

Similarly, the first test `test_single_contributor_one_contribution` makes an update to an episode and checks that contributions were properly registered, which is the functionality we're testing here. Then the test also checks that the episode was renamed, but this shouldn't be necessary. If we want to check that we can rename episodes with `updateEpisode`, there should be a separate unit test in the `event` app for that. In this test, you should be able to trust that `updateEpisode` is properly tested, and you only need to test the contributions.

backend/core/tests/test_contributors.py
```diff
@@ -0,0 +1,239 @@
...
+        self.assertEqual(len(content["data"]["updateEpisode"]["errors"]), 0)
+
+        episode = Episode.objects.get(id=self.episode.pk)
+
+        self.assertEqual(episode.name, "Darth Vader")
+        self.assertEqual(
+            len(episode.contributors.all()),
+            1,
+            "Contributors should not increase after multiple contributions.",
+        )
```
lukavdplas, line 101: Since `episode.contributors` is a many-to-many field, you can't get duplicates. To be honest, I don't find it necessary to test this, since we can trust that Django (or PostgreSQL) has taken care of that; it's not something we're handling in our own code.

That said,  there is little harm in superfluous testing, so this isn't a major issue.

lukavdplas (approved):
Neat!

I added some comments about the unit tests, but those are not hard requirements.

XanderVertegaal:
Thanks for the constructive comments! I really appreciate them :)

I wasn't 100% sure about M2M fields not containing which is why I wrote the tests. Now that I know, I won't bother testing this behaviour. You're right that we can trust Postgres/Django on this.

Your point on keeping tests simple and focused is well taken. I've simplified the ones in this PR for now and will take care not to test unrelated properties in future PRs.

PR #120, opened 2024-09-03
Feature/item gift form

XanderVertegaal:
Implements forms with (non-M2M) fields for Gifts and Letters, based on the structure in #113 (It makes sense to review that one first.)

Update, Create and Delete mutations have been added, although only the Update mutation is currently used. (The other two require modals, which are not part of this PR.)


frontend/src/app/data-entry/gift-form/gift-categories-form/gift-categories-form.component.ts
```diff
@@ -0,0 +1,54 @@
+import { Component, DestroyRef } from "@angular/core";
```
XanderVertegaal, line 1: Just a few imports/properties we might need in the future. These are not used yet.

frontend/src/app/data-entry/gift-form/gift-identification-form/gift-identification-form.component.ts
```diff
@@ -0,0 +1,105 @@
...
+                    emitEvent: false,
+                    onlySelf: true,
+                });
+            });
+
+        this.gift$
+            .pipe(
+                switchMap(() =>
+                    this.form.valueChanges.pipe(
+                        map(() => this.form.getRawValue()),
```
lukavdplas, line 68: What does this do? I though this was already the payload of `form.valueChanges`?

frontend/src/app/data-entry/gift-form/gift-identification-form/gift-identification-form.component.ts
```diff
@@ -0,0 +1,105 @@
...
+                                {
+                                    update: (cache) => {
+                                        cache.evict({
+                                            id: `GiftDescriptionType:${id}`,
+                                        });
+                                        cache.gc();
+                                    },
+                                }
+                            )
+                        )
```
lukavdplas, line 89: For the sake of readability, please outfactor the `this.giftMutation.mutate` call to its own method.

lukavdplas (approved):

frontend/src/app/data-entry/gift-form/gift-identification-form/gift-identification-form.component.ts
XanderVertegaal: By default, you can disable any control in a FormGroup and their value won't show up in `form.value` (which is passed by `form.valueChanges`). As a result, `form.value`  is typed with `Partial<...>` to reflect the fact that all properties are optional. `.getRawValue()` takes the value of all controls, disabled or not, so the value has the type expected by the mutation.

frontend/src/app/data-entry/gift-form/gift-identification-form/gift-identification-form.component.ts
XanderVertegaal: Done!

PR #126, opened 2024-09-19
Episode-entity relationship form

### LINKED ISSUE #125: Episode-entity relation form
This is currently in progress, I'm creating the issue for reference.

The data entry form should include a menu to edit the agents, locations, gifts, and letters involved in an episode. (And vice versa, the episodes in which an agent/location/gift/letter is involved.)

These menus are very similar, so they can rely on the same underlying component. Also, some of the UI may also be reused in other many-to-many relationships, such as #124

lukavdplas:
This adds forms for:

- Editing related episodes on the agent form
- Editing related agents, locations, letters, and gifts on the episode form

Close #125 

Includes:
- Some mutations and a query to handle the relationship between an episode and a related entity (i.e. an agent/location/gift/letter). Also adds some interfaces to work with shared properties.
- GraphQL types are now explicit that `name` and `description` are non-nullable in the query output.
- A `EpisodeEntitiesComponent` to be used in the episode form for agents, locations, gifts, and letters.
- An `AgentEpisodesComponent` to be used in the agent form. This one is not directly reusable (yet).
- An `EpisodeLinkComponent` and `CollapsibleCardComponent` as more basic building blocks.

Not (yet) included:
- You can't yet edit related episodes on the forms for locations, letters, and gifts. This was harder to generalise than the other way around, but could still be done in the future. Or we could just copy the `AgentEpisodesComponent`.
- The dropdown to add a new item includes a "create" button, but it's not yet implemented.


backend/event/types/EpisodeAgentType.py
```diff
@@ -1,24 +1,23 @@
...
+    entity_type = NonNull(Entity)
+
     class Meta:
-        model = EpisodeGift
+        model = EpisodeAgent
         fields = [
             "id",
             "episode",
             "agent",
+            "entity",
```
XanderVertegaal, line 20: "entity" and "entity_type" do not exist on the model, so they don't belong here.

frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.ts
```diff
@@ -0,0 +1,160 @@
...
+    addEntity(entityID: string, episodeID: string): void {
+        const input: CreateEpisodeEntityLinkInput = {
+            entity: entityID,
+            episode: episodeID,
+            entityType: this.entityType,
+        };
+        this.addMutation.mutate({ input }, {
+            refetchQueries: REFETCH_QUERIES,
+        }).pipe(
+            tap(() => this.status$.next('loading'))
```
XanderVertegaal, line 121: This `tap()` will only run when the result of the mutation is already in.

frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.ts
```diff
@@ -0,0 +1,160 @@
...
+
+let nextID = 0;
+
+@Component({
+    selector: "lc-episode-entities-form",
+    templateUrl: "./episode-entities-form.component.html",
+    styleUrls: ["./episode-entities-form.component.scss"],
+})
+export class EpisodeEntitiesFormComponent implements OnChanges, OnDestroy {
+    @Input() entityType: Entity = Entity.Agent;
```
XanderVertegaal, line 31: Why would this be the default? Maybe rather `@Input({required = true}) entityType!: Entity;` (or `... entitytype: Entity | null = null`, if you prefer).

frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    private availableEntities(
+        data: DataEntryEpisodeEntitiesQuery
+    ): { name: string, id: string }[] {
+        const allEntities: {
+            __typename?: string,
+            name: string,
+            id: string
+        }[] = data.episode?.source[this.entityProperty] || [];
+        const linkedEntities = data.episode?.[this.entityProperty] || [];
```
XanderVertegaal, line 157:
```suggestion
        const linkedEntities = this.linkedEntities(data);
```

frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    removeEntity(entityID: string, episodeID: string): void {
+        const data: DataEntryDeleteEpisodeEntityLinkMutationVariables = {
+            entity: entityID,
+            episode: episodeID,
+            entityType: this.entityType,
+        };
+        this.removeMutation.mutate(data, { refetchQueries: REFETCH_QUERIES }).pipe(
+            tap(() => this.status$.next('loading'))
```
XanderVertegaal, line 132: See comment above.

frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.ts
```diff
@@ -0,0 +1,160 @@
...
+    linkedEntities$: Observable<{ id: string, name: string }[]>;
+    availableEntities$: Observable<{ name: string, id: string }[]>;
+
+    addEntity$ = new Subject<string>();
+    removeEntity$ = new Subject<string>();
+    actionIcons = actionIcons;
+    status$ = formStatusSubject();
+    id = `episode-entities-${nextID++}`;
+
+    private mutationObserver: Partial<Observer<MutationResult>> = {
```
XanderVertegaal, line 44: This confused me because of the [Web API with the same name](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver).

frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.ts
```diff
@@ -0,0 +1,160 @@
...
+    DataEntryDeleteEpisodeEntityLinkGQL,
+    DataEntryDeleteEpisodeEntityLinkMutationVariables,
+    DataEntryEpisodeEntitiesGQL,
+    DataEntryEpisodeEntitiesQuery,
+    Entity
+} from "generated/graphql";
+import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
+import { splat, differenceBy } from "@shared/utils";
+
+type EntityPropertyName = 'agents' | 'gifts' | 'letters' | 'spaces';
```
XanderVertegaal, line 19: Or:

```suggestion
type EntityPropertyName = `${Lowercase<keyof typeof Entity>}s`;
```

frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    data$: Observable<DataEntryEpisodeEntitiesQuery>;
+
+    linkedEntities$: Observable<{ id: string, name: string }[]>;
+    availableEntities$: Observable<{ name: string, id: string }[]>;
+
+    addEntity$ = new Subject<string>();
+    removeEntity$ = new Subject<string>();
+    actionIcons = actionIcons;
+    status$ = formStatusSubject();
```
XanderVertegaal, line 41: For clarity reasons, I'm not sure I prefer this outfactor to the single line it replaces, but it works.

frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.ts
```diff
@@ -0,0 +1,160 @@
+import { Component, ElementRef, Input, OnChanges, OnDestroy, SimpleChanges, ViewChild } from "@angular/core";
```
XanderVertegaal, line 1: Unused imports.

frontend/src/app/data-entry/episode-form/episode-form.component.html
```diff
@@ -21,13 +21,17 @@ <h2 class="subform-header">Contents</h2>
 <lc-episode-contents-form />
 
 <h2 class="subform-header">Agents</h2>
-<lc-episode-agents-form />
+<lc-episode-entities-form [entityType]="Entity.Agent" />
```
XanderVertegaal, line 24: Very neat!

frontend/src/app/data-entry/episode-form/episode-form.component.ts
```diff
@@ -1,15 +1,19 @@
...
-import { DataEntryEpisodeFormGQL } from "generated/graphql";
+import { DataEntryEpisodeFormGQL, Entity } from "generated/graphql";
 import { filter, map, share, switchMap } from "rxjs";
+import { FormService } from "../shared/form.service";
 
 @Component({
     selector: "lc-episode",
     templateUrl: "./episode-form.component.html",
     styleUrls: ["./episode-form.component.scss"],
+    providers: [FormService],
```
XanderVertegaal, line 12: In an upcoming PR I move this provider line from the individual components to the module. We don't need multiple instantiations of FormService, and it breaks the singleton structure.

frontend/src/app/data-entry/gift-form/gift-identification-form/gift-identification-form.component.ts
```diff
@@ -7,8 +7,10 @@ import { ToastService } from "@services/toast.service";
 import { MutationResult } from "apollo-angular";
 import {
     DataEntryGiftIdentificationGQL,
+    DataEntryGiftIdentificationQuery,
```
XanderVertegaal, line 10: This is not used.

frontend/src/app/data-entry/shared/types.ts
```diff
@@ -0,0 +1,5 @@
+export type FormStatus = 'idle' | 'invalid' | 'loading' | 'saved' | 'error';
+
+export type EntityType = 'agent';
```
XanderVertegaal, line 3: This type does not appear to be used.

frontend/src/app/data-entry/agent-form/agent-episodes-form/agent-episodes-form.component.ts
```diff
@@ -0,0 +1,119 @@
...
+        this.removeEpisode$.pipe(
+            withLatestFrom(this.formService.id$),
+            takeUntilDestroyed(),
+        ).subscribe(([episodeID, agentID]) =>
+            this.removeEpisode(episodeID, agentID)
+        );
+    }
+
+    ngOnDestroy(): void {
+        this.status$.complete();
```
XanderVertegaal, line 73: With `takeUntilDestroyed()` I don't think you need to complete these observables.

frontend/src/app/data-entry/agent-form/agent-episodes-form/agent-episodes-form.component.ts
```diff
@@ -0,0 +1,119 @@
...
+    addEpisode(episodeID: string, agentID: string): void {
+        const input: CreateEpisodeEntityLinkInput = {
+            entity: agentID,
+            episode: episodeID,
+            entityType: Entity.Agent,
+        };
+        this.addMutation.mutate({ input }, {
+            refetchQueries: REFETCH_QUERIES,
+        }).pipe(
+            tap(() => this.status$.next('loading'))
```
XanderVertegaal, line 88: This `tap()` kicks in only after the mutation is completed.

frontend/src/app/data-entry/agent-form/agent-episodes-form/agent-episodes-form.component.html
```diff
@@ -0,0 +1,40 @@
+<ng-container *ngIf="data$ | async as data">
+    <ng-container *ngIf="data.agentDescription">
```
XanderVertegaal, line 2: You can merge these `ng-container`s:

```html
    <ng-container *ngIf="(data$ | async).agentDescription as agent">
```

frontend/src/app/data-entry/shared/episode-link-form/episode-link-form.component.ts
```diff
@@ -0,0 +1,174 @@
...
+    get notesInputID() {
+        return this.id + '-notes-input';
+    }
+
+    get entityName(): string {
+        return entityTypeNames[this.entityType];
+    }
+
+    get linkedObjectName(): string {
+        return this.linkTo == 'episode' ? 'episode' : this.entityName;
```
XanderVertegaal, line 101:
```suggestion
        return this.linkTo == 'episode' ?? this.entityName;
```

(All of these getters would be sweet `computed()`s by the way, but that's something to consider when we're less strapped for time.)

frontend/src/app/data-entry/shared/episode-link-form/episode-link-form.component.ts
```diff
@@ -0,0 +1,174 @@
...
+    sourceMentionOptions: SelectOptions<SourceMention> = sourceMentionSelectOptions();
+
+    private link$ = new Subject<UpdateEpisodeEntityLinkInput>();
+    private status$ = formStatusSubject();
+    private id = `episode-link-${nextID++}`;
+
+    actionIcons = actionIcons;
+
+    constructor(
+        query: DataEntryEpisodeEntityLinkGQL,
```
XanderVertegaal, line 62: `private`? Also a better name than `query` would be helpful.

frontend/src/app/data-entry/shared/episode-link-form/episode-link-form.component.ts
```diff
@@ -0,0 +1,174 @@
...
+        this.link$.complete();
+        this.status$.complete();
+        this.formService.detachForm(this.id);
+    }
+
+    formUrl(linkTo: LinkTo, entityType: Entity): string {
+        if (linkTo == 'episode') {
+            return 'episodes';
+        }
+        const routes = {
```
XanderVertegaal, line 124: This object is very similar to `entityTypeNames` in `utils.ts`. For consistency's sake, please put this object there as well. (You can maybe merge them as well, or derive one from the other). If you've done that, you can replace this function's body with `return linkTo == 'episode' ? 'episodes' : entityTypeRoutes[entityType];`

frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.html
```diff
@@ -0,0 +1,48 @@
...
+            </p>
+        </ng-template>
+    </ng-container>
+</ng-container>
+
+<div class="mt-4 mb-4">
+    <div ngbDropdown>
+        <button type="button" class="btn btn-primary"
+            #dropdownToggle
+            [attr.id]="dropdownToggleID"
```
XanderVertegaal, line 32: Is this diffferent from `[id]` you use elsewhere?

XanderVertegaal (commented):
Very impressive how you managed to  implement this despite the sparsity of the documentation. Well done!

I have a lot of nit-picky comments, but functionally, the code works great! Consider it approved after you've covered my remarks.

frontend/src/app/data-entry/episode-form/episode-form.component.ts
lukavdplas: Conceptually, the service is used so the component can aggregate the status of its child components. If you were to create multiple instances of `EpisodeFormComponent` (or any of the other forms that provide a FormService), you would absolutely need multiple instances of the service, too.

Angular services are often stateless or keep track of a global state in the application, which makes them work best as singletons, but a singleton structure is not inherently superior; in this case, the service is tracking the local state of the component, so making it a singleton sounds like it invites bugs in my opinion.

frontend/src/app/data-entry/shared/utils.ts
```diff
@@ -0,0 +1,19 @@
+import { Entity, SourceMention } from "generated/graphql";
+import { FormStatus, SelectOptions } from "./types";
+import { BehaviorSubject } from "rxjs";
+import _ from "underscore";
+
+export const sourceMentionSelectOptions = (): SelectOptions<SourceMention> => [
```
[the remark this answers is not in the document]
lukavdplas: It's not really necessary, but as a general rule, it's a good idea to only use global constants with immutable types. You can get side effects if you use an array as a global constant.

frontend/src/app/data-entry/episode-form/episode-form.component.ts
XanderVertegaal: I did not intend to suggest that the singleton pattern is superior, but I see your point. I'll adjust my PR.

frontend/src/app/shared/utils.ts
```diff
@@ -0,0 +1,37 @@
+import _ from "underscore";
+
+/**
+ * Exclude all values from a list that have a matching value in another list, based
+ * on the specified property.
+ *
+ * Similar to _.difference, but compares on a specified key.
+ *
+ * @param list The starting list
+ * @param others The list with values to excluded
```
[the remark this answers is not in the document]
lukavdplas: In general, I would say it's a good principle to have an explicit "translation" function between internal variable names and interface content; even in cases when you could technically get away with something like `toLowercase`.

- It makes the code more suitable for localisation.
- Normally, programmers assume that symbol names are arbitrary, and we can rename them at our convenience.
- You want to be somewhat flexible with phrasing (such as the space/place/location question); you should be able to edit the text without needing to update the entire application.

But that's also a balance between following a general principle and doing whatever is easiest in this particular case.

frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.html
lukavdplas: No. `id` is one of the few HTML attributes that Angular supports without `attr.`, but because most HTML attributes require `attr.*`, I sometimes type that out of habit.

PR #127, opened 2024-09-19
Miscellaneous form fixes

lukavdplas:
This include a bunch of small fixes and improvements I made when getting the application ready for deployment.

Changes:
- Some of the text in the episode preview is cleared up, e.g. "no letters added yet" -> "no letters added" (as the episode may not have any letters).
- There was a confusing overlap between the "description" and "summary" fields for episodes. I removed descriptions from the form.
- Episode previews also display the page, instead of just book/chapter.
- Added proper cache update function to several mutations in the frontend.
- Show proper status updates in the episode form.
- Fixed some compatability issues for Python 3.9 (the version used on the server), and changed the Python version in the test workflow to 3.9.
- Added contributors in some mutations that were missing that.
- Use the "group" instead of "person" icon in the header of the agent form when appropriate.
- Increased the budget for building the frontend.
- Added an alert when the mutations for adding/removing entities fail.

### DIFF
--- a/backend/event/mutations/CreateEpisodeEntityLinkMutation.py
+++ b/backend/event/mutations/CreateEpisodeEntityLinkMutation.py
@@ -81,6 +81,7 @@ class CreateEpisodeEntityLinkMutation(LettercraftMutation):
         data: CreateEpisodeEntityLinkInput,
         info: ResolveInfo,
     ):
-        user = info.context.user
-        episode_entity_link.episode.contributors.add(user)
-        episode_entity_link.entity.contributors.add(user)
+        if info.context:
+            user = info.context.user
+            episode_entity_link.episode.contributors.add(user)
+            episode_entity_link.entity.contributors.add(user)
--- a/backend/event/mutations/CreateEpisodeEntityLinkMutation.py
+++ b/backend/event/mutations/CreateEpisodeEntityLinkMutation.py
@@ -65,6 +65,7 @@ class CreateEpisodeEntityLinkMutation(LettercraftMutation):
                 obj = LinkModel.objects.create(
                     **{"episode": episode, LinkModel.entity_field: entity}
                 )
+                cls.add_contribution(obj, data, info)
                 obj.clean()
         except ValidationError as e:
             errors = [
@@ -74,3 +75,12 @@ class CreateEpisodeEntityLinkMutation(LettercraftMutation):
             return cls(ok=False, errors=errors)
 
         return cls(ok=True, errors=[])
+
+    def add_contribution(
+        episode_entity_link: EpisodeEntity,
+        data: CreateEpisodeEntityLinkInput,
+        info: ResolveInfo,
+    ):
+        user = info.context.user
+        episode_entity_link.episode.contributors.add(user)
+        episode_entity_link.entity.contributors.add(user)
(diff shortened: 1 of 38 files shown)


frontend/src/app/services/modal.service.ts
```diff
@@ -13,7 +13,7 @@ interface ConfirmationModalConfig {
 export class ModalService {
     constructor(private modal: NgbModal) {}
 
-    public openConfirmationModal(config: ConfirmationModalConfig): Promise<void> {
+    public openConfirmationModal(config: ConfirmationModalConfig): Promise<any> {
```
XanderVertegaal, line 16: Why `any`? We know that this modal won't return anything.

frontend/src/app/data-entry/episode-form/episode-identification-form/episode-identification-form.component.ts
```diff
@@ -120,12 +116,26 @@ export class EpisodeIdentificationFormComponent implements OnInit {
         );
     }
 
-    private updateCache(cache: ApolloCache<unknown>, id: string): void {
+    private updateCache(cache: any, id: string): void {
```
XanderVertegaal, line 119:
```suggestion
import { ApolloCache } from "@apollo/client/core";

    private updateCache(cache: ApolloCache<unknown>, id: string): void {
```

XanderVertegaal (approved):
Thanks for making these changes! They are all very sensible, and I fully support suppressing the `description` field on `episode` (given that we already have a `summary`) there as well.

I have a few minor tidbits for your reading pleasure, but nothing critical.

backend/event/mutations/CreateEpisodeEntityLinkMutation.py
```diff
@@ -74,3 +75,13 @@ def mutate(cls, root: None, info: ResolveInfo, data: CreateEpisodeEntityLinkInpu
...
+
+    def add_contribution(
+        episode_entity_link: EpisodeEntity,
+        data: CreateEpisodeEntityLinkInput,
+        info: ResolveInfo,
+    ):
+        if info.context:
+            user = info.context.user
+            episode_entity_link.episode.contributors.add(user)
+            episode_entity_link.entity.contributors.add(user)
```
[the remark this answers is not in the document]
lukavdplas: > I don't think I've ever encountered a situation in which info.context was falsy. Have you run into this? As far as I can see, the source code does not suggest that it's something we should have to worry about.

Yes, this happens during unit tests (and was causing them to fail). But yes, as far as I know, "real" requests always have a context.

> I am a bit worried about AnonymousUser, however. It should not be a problem because the forms are hidden behind a login, but technically, I guess we could have AnonymousUser as a contributor? Perhaps we should check whether user.is_authenticated(); what do you think?

Hm, I can't imagine we would ever allow anonymous users to make contributions; when we open up the site, anonymous users should have read-only access. So I think it's fine if we don't cover that case here.

`AnonymousUser` is not a database object, by the way (and it's not a `User` either). So if an anonymous request somehow went through, `contributors.add(user)` would raise an error. This would mean the entire transaction is rolled back, and the user receives an error response. That seems fine to me.

If you wanted to allow anonymous contributions, you would need to adjust the database models to store those somehow. But like I said, I don't think this is going to be relevant.

PR #135, opened 2024-09-26
Bugfix/nameless contributors

lukavdplas:
- More sensible display names when a user has not filled in their (complete) name
- Clears the cached name of the user when the update their name in the settings menu.

### DIFF
--- a/frontend/src/app/user/user-settings/user-settings.component.ts
+++ b/frontend/src/app/user/user-settings/user-settings.component.ts
@@ -1,7 +1,7 @@
 import { Component, DestroyRef, OnInit } from "@angular/core";
 import { FormControl, FormGroup, Validators } from "@angular/forms";
 import { AuthService } from "@services/auth.service";
-import { UserSettings } from "../models/user";
+import { UserResponse, UserSettings } from "../models/user";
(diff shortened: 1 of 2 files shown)


frontend/src/app/user/user-settings/user-settings.component.ts
```diff
@@ -132,4 +130,22 @@ export class UserSettingsComponent implements OnInit {
...
+        });
+        this.updateCache(user.id);
+    }
+
+    private updateCache(id: number) {
+        const cache = this.apollo.client.cache;
+        const identified = cache.identify({ __typename: 'UserType', id });
+        cache.evict({ id: identified, fieldName: 'firstName' });
+        cache.evict({ id: identified, fieldName: 'lastName' });
+        cache.evict({ id: identified, fieldName: 'fullName' });
```
XanderVertegaal, line 148: Is there any reason why you specifically want to evict these fields and not do `cache.evict({ id: identified })` in general?

lukavdplas: It's more precise because those are the only fields that are edited here. But I think those are also the only fields, so this is not really necessary. Setting a `fieldName` is more useful with the data entry forms, where evicting specific fields really avoids a lot of needless requests.

PR #137, opened 2024-09-26
update text fields on blur

### LINKED ISSUE #129: Backend sync interrupts text field input
Data in forms is automatically saved to the backend when the user makes changes. After saving, the form input is updated with the data from the backend. This can create an annoying experience if the user is typing in a text field while the autosave happens:

- The contents of the text fields are updated from the backend, so whatever the user typed in the last second or so is lost.
- The position of the cursor jumps to the end of the text input.

Possible solutions:

- Change the autosave so it's triggered when the form loses focus, instead of on changes. This is fairly easy to implement and should more-or-less solve the issue. (You would still get the same issue in some cases, e.g. if you click outside of the text field, which triggers the sync, then immediately click inside it and start typing. But this isn't such a common scenario.)
- When you're updating the values of form fields based on the backend, and an input element has focus, then arguably that field shouldn't be updated at all. Whatever the user is typing should be given precedence over the data from the backend.

lukavdplas:
Quick fix for the autosaving issue (close #129)

Changes the configuration of fields with free text entry so they update on blur instead of on changes; that is, changes will only be propagated when the user moves focus out of the field.

The setting is only used with fields that are auto-saving. Text fields in modals, which have a submit action, don't need this.


lukavdplas:
> I also ran into the issue that with updateOn: 'blur', the user input won't be saved when the user navigates to a different page while the input still has focus. 

Yes, I had considered this setting earlier and dismissed it for that reason. Still, this is definitely better than the current situation.

> You could see if you can solve this by triggering one final form submission in ngOnDestroy.

I'm not too keen on that solution because it would mean components have to start asynchronous tasks when they're being destroyed. Also, keep in mind that `ngOnDestroy` only works if you navigate within the application, not if you refresh the page, close the browser, go to a different site, or go to the admin interface.

XanderVertegaal:
> Yes, I had considered this setting earlier and dismissed it for that reason. Still, this is definitely better than the current situation.

It definitely is! 👍 

> I'm not too keen on that solution because it would mean components have to start asynchronous tasks when they're being destroyed. Also, keep in mind that `ngOnDestroy` only works if you navigate within the application, not if you refresh the page, close the browser, go to a different site, or go to the admin interface.

There are [ways](https://ultimatecourses.com/blog/exploring-angular-lifecycle-hooks-ondestroy#making-ngondestroy-async) to make this work, apparently, but I understand it if this is not the way you want to go.

PR #139, opened 2024-09-27
Ordering and icons for agents + locations

### LINKED ISSUE #88: Sensible order for episode agents
The overview of an episode in the data entry interface should lists agents in a "sensible" order. Suggested order: historical figures first, then unknown individuals, then groups.

lukavdplas:
- Adjusts the ordering for agents and locations so identified objects (i.e. tied to a historical person / location) appear first. For agents, people also appear before groups. close #88 
- Use the filled / outline styles for location icons, based on whether the location has identifiable features.
- For both agents and locations, there is now a boolean property in the graphQL API to check whether they are identified.


backend/person/models.py
```diff
@@ -74,6 +74,12 @@ class AgentDescription(EntityDescription, models.Model):
     A description of an agent in a source text; can be a single person or a group
     """
 
+    class Meta:
+        ordering = [
+            models.F("describes__identifiable").desc(nulls_last=True),
```
XanderVertegaal, line 79: I think you forgot to make new migrations.

backend/person/models.py
```diff
@@ -93,6 +99,9 @@ def clean(self):
                 "Only groups can describe multiple historical figures"
             )
 
+    def identified(self):
```
XanderVertegaal, line 102: Would it be an idea to decorate this with `@property`?

backend/space/models.py
```diff
@@ -28,6 +38,13 @@ class SpaceDescription(EntityDescription, models.Model):
         help_text="Man-made structures referenced in this description",
     )
 
+    def has_identifiable_features(self):
```
XanderVertegaal, line 41: This could likewise be a `@property`.

XanderVertegaal (changes requested):
If you can add the migrations, you can consider this approved.

PR #141, opened 2024-10-01
Basic location form

XanderVertegaal:
Just the Identification and Source Text subforms, no M2M-forms for Region/Settlement/Structure yet.

This does not yet implement a delete button (to be added after #128).


frontend/src/app/data-entry/location-form/location-source-text-form/location-source-text-form.component.ts
```diff
@@ -0,0 +1,135 @@
...
+            });
+
+        this.form.statusChanges
+            .pipe(
+                filter((status) => status == "INVALID"),
+                takeUntilDestroyed(this.destroyRef)
+            )
+            .subscribe(() => this.status$.next("invalid"));
+
+        const validFormSubmission$ = this.location$.pipe(
```
lukavdplas, line 86: It's not clear to my why this is piped on `this.location$` - since looks like you're immediately switching to `() => this.form.valueChanges`, what do you need the location observable for?

frontend/src/app/data-entry/location-form/location-form.component.ts
```diff
@@ -1,55 +1,128 @@
...
+                        header: "Deletion failed",
+                    });
+                } else {
+                    this.toastService.show({
+                        body: "Location deleted",
+                        type: "success",
+                        header: "Success",
+                    });
+                }
+                this.router.navigate([`/data-entry/sources/${sourceId}`]);
```
lukavdplas, line 116: For long nested functions like this, I find it more readable if they're moved into their own function/method.

lukavdplas (changes requested):
Neat! Only minor suggestions.

frontend/src/app/data-entry/location-form/location-form.component.ts
XanderVertegaal: I agree; again this was copied from a branch that did not yet have all the outfactors to separate functions.

frontend/src/app/data-entry/location-form/location-identification-form/location-identification-form.component.html
```diff
@@ -0,0 +1,31 @@
+<form>
+    <div class="form-group">
+        <label class="form-label" for="episode-name">Name</label>
+        <p class="form-text">
+            This is a name to identify the location in the overview of the
+            source. For example: "Kingdom of Burgundy", "Abbey of St Gall".
```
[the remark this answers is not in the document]
XanderVertegaal: Nice! Done :)

frontend/src/app/data-entry/location-form/location-source-text-form/location-source-text-form.component.ts
XanderVertegaal: I wanted to only start firing mutations on `form.valueChanges` after the initial data has been loaded in. My idea was that this way, the user can't really change anything if the internet connection is slow and the data is not fully loaded in yet. Originally, it was meant to prevent the initial load from triggering a mutation, but `emitEvent: false` and `onlySelf: true` in `location$.subscribe` also accomplish this.

frontend/src/app/data-entry/location-form/location-source-text-form/location-source-text-form.component.ts
```diff
@@ -0,0 +1,135 @@
...
+            switchMap(() =>
+                this.form.valueChanges.pipe(
+                    map(() => this.form.getRawValue()),
+                    filter(() => this.form.valid)
+                )
+            ),
```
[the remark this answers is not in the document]
XanderVertegaal: I guess that by performing the mapping/filtering in the outer observable, the result of these operations is not ignored when a new emission comes in at the same time. By keeping them in the inner Observable, the mapping/filtering result is properly discarded when a new emission comes in.

That said, this is a theoretical more than anything. For readability's sake I'll refactor this.

lukavdplas (approved):

PR #142, opened 2024-10-02
Feature/user permissions

### LINKED ISSUE #138: Permission for contributors
Currently, all registered, non-admin users have the same privileges. We should implement a permission for contributors, i.e. users with the right to enter research data through the frontend.

lukavdplas:
Users now need to be given explicit permission to contribute research data. Close #138 

- Adds an "is contributor" option in the admin interface for users
- Authentication is no longer required to use the graphQL API. However, only contributors are allowed to make mutations.
- Only contributors are allowed to visit the data entry pages in the frontend.
- The `LoggedOnGuard` for frontend routes is now also used in the user settings page, and shows a message when the user is redirected.


frontend/src/app/shared/logged-on.guard.ts
```diff
@@ -1,20 +1,52 @@
...
+                return router.createUrlTree(['/login'], {
+                    queryParams: { next: route.url }
+                });
             }
             return true;
         })
     );
 };
+
+export const ContributorGuard: CanActivateFn = (route: ActivatedRouteSnapshot) => {
```
XanderVertegaal, line 30: You might want to move this to a separate file (or change the name of this file).

frontend/src/app/shared/logged-on.guard.ts
```diff
@@ -1,20 +1,52 @@
...
 
 export const LoggedOnGuard: CanActivateFn = (route: ActivatedRouteSnapshot) => {
     const authService = inject(AuthService);
+    const toastService = inject(ToastService);
     const router = inject(Router);
 
     return authService.currentUser$.pipe(
         filter((user) => user !== undefined),
         map((user) => {
+            toastService.show({
```
XanderVertegaal, line 15: You need to move `if (user === null) {` or some other check here. Currently, the toast is shown whenever this guard is triggered.

XanderVertegaal (changes requested):
Nice! Feel free to merge after addressing the two small comments I added.

PR #154, opened 2024-10-25
Feature/episode order

### LINKED ISSUE #122: Sort episodes by book/chapter/page
The order of episodes is now based on ID (by default), which is not always logical.

XanderVertegaal:
Closes #122 

Adds a `rank` property to objects inheriting EntityDescription and allows the user to drag/drop episodes in a source's episode overview to reorder them (using Angular Material's CDK).

I'd be happy to discuss alternatives if you think rank-based ordering or a drag-and-drop interface are not the way to go.


XanderVertegaal:
This PR will be adjusted to replace the drag and drop interface with buttons to move episodes 'up'/'down' in the hierarchy.

backend/event/mutations/UpdateEpisodeOrderMutation.py
```diff
@@ -0,0 +1,39 @@
...
+
+    class Arguments:
+        episode_order_data = List(NonNull(EpisodeOrderInput), required=True)
+
+    @classmethod
+    def mutate(cls, root: None, info: ResolveInfo, episode_order_data: list[EpisodeOrderInput]):
+        # Map Episode IDs to new rank values
+        id_to_rank: dict[int, int] = {
+            int(episode.id): episode.rank for episode in episode_order_data
+        }
```
lukavdplas, line 25: I think these type declarations need to be `List` and `Dict` instead of `list` and `dict`, right?

lukavdplas (changes requested):
For clarity, the point about the drag-and-drop UI was something we discussed outside of Github.

I added some additional comments on the backend. These are smaller suggestions, but please take a look at those as well.

XanderVertegaal:
I will reopen another PR with `order_with_respect_to` and with a new interface.

PR #155, opened 2024-10-30
Feature/order episodes in source

### LINKED ISSUE #122: Sort episodes by book/chapter/page

XanderVertegaal:
Closes #122 

Implements `order_with_respect_to` to keep track of the order of episodes within a source and adds buttons in the frontend so users can manually adjust the order.


frontend/src/app/data-entry/source/source.component.ts
```diff
@@ -75,4 +88,65 @@ export class SourceComponent {
...
+
+        const indexNotFound = currentIndex === -1;
+        const indexAtBoundary =
+            (change === "up" && currentIndex <= 0) ||
+            (change === "down" && currentIndex >= episodeIds.length - 1);
+
+        // Don't mutate if the order change is invalid.
+        if (indexNotFound || indexAtBoundary) {
+            return;
+        }
```
lukavdplas, line 108: These cases should be avoided by disabling the button.

frontend/src/app/shared/utils.ts
```diff
@@ -35,3 +35,29 @@ export const differenceBy = <T extends object>(
...
+    if (fromIndex === toIndex) {
+        return array;
+    }
+
+    const fromIndexOutOfBounds = fromIndex < 0 || fromIndex >= array.length;
+    const toIndexOutOfBounds = toIndex < 0 || toIndex >= array.length;
+
+    if (fromIndexOutOfBounds || toIndexOutOfBounds) {
+        return array;
+    }
```
lukavdplas, line 58: These situations are real mistakes, I would expect the function to throw an error here.

frontend/src/app/shared/order-button-group/order-button-group.component.ts
```diff
@@ -0,0 +1,24 @@
...
+export type OrderChange = "up" | "down";
+
+@Component({
+    selector: "lc-order-button-group",
+    templateUrl: "./order-button-group.component.html",
+    styleUrls: ["./order-button-group.component.scss"],
+})
+export class OrderButtonGroupComponent {
+    @Input() entityName: string | null = null;
+    @Output() changeOrder = new EventEmitter<OrderChange>();
```
lukavdplas, line 13: You could also just emit `1` or `-1` here, which makes handling the event easier. That's a matter of personal preference, though.

frontend/src/app/shared/order-button-group/order-button-group.component.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export type OrderChange = "up" | "down";
+
+@Component({
+    selector: "lc-order-button-group",
+    templateUrl: "./order-button-group.component.html",
+    styleUrls: ["./order-button-group.component.scss"],
+})
+export class OrderButtonGroupComponent {
+    @Input() entityName: string | null = null;
```
lukavdplas, line 12: As far as I know, there is no reason to make this configurable, at least in the foreseeable future. You can just write "episode" in the template.

frontend/src/app/shared/icons.ts
```diff
@@ -18,6 +18,8 @@ export const actionIcons = {
     expand: 'caret-down',
     collapse: 'caret-up',
     edit: 'pencil',
+    up: 'chevron-up',
+    down: 'chevron-down',
```
lukavdplas, line 22: Hm, this might look too similar to the `collapse`/`expand` icons, which appear in a similar position. Perhaps `'arrow-up'` / `'arrow-down'` would be more distinctive?

lukavdplas (changes requested):
Nice!

This looks good overall; I've added some requests for minor changes.

frontend/src/app/shared/utils.ts
XanderVertegaal: Good idea!

backend/event/mutations/UpdateEpisodeOrderMutation.py
```diff
@@ -15,17 +15,30 @@ class Arguments:
...
-        if source.count() > 1:
+        corresponding_sources = Source.objects.filter(episode__in=episode_ids)
+        if corresponding_sources.count() != len(episode_ids):
             error = LettercraftErrorType(
                 field="episode_ids",
-                messages=["Multiple sources found for given episode IDs."],
+                messages=["Not every episode has a corresponding source."]
             )
-            return cls(ok=False, errors=[error])  # type: ignore
+            return cls(ok=False, errors=[error]) # type: ignore
```
lukavdplas, line 31: This check is hard to understand without a comment since it relies on Django returning duplicate rows for the `filter` query, which is kind of obscure knowledge. It's also an indirect test for the thing being validated (invalid IDs), which also makes it less intuitive.

The message here is also confusing. Strictly speaking, every episode *does* have a corresponding source, but not every ID has a corresponding episode. Practically, the feedback to the client should be that it's sending non-existent episode IDs, not that there is an issue in the `source` field of episodes.

lukavdplas (changes requested):

backend/event/mutations/UpdateEpisodeOrderMutation.py
XanderVertegaal: Wow, so basically everything here is bad then 😅

I've made some changes in my last commit, but if this is still not how you want it to be, feel free to push your own commits.

lukavdplas: I mean, it was doing the right thing, just confusing to read. Your updated version looks fine!

PR #157, opened 2024-11-22
Feature: delete user

### LINKED ISSUE #66: Delete User
A view has been added for this in user/views.py, and the frontend has a button in place that makes a call to its associated endpoint, but currently nothing happens.

XanderVertegaal:
Closes #66.

This PR allows users to delete their own accounts (in the user settings menu). A pop-up is shown to confirm before deletion actually takes place.

Strangely, user deletion with `django-allauth` does not work without the `socialaccount` app installed, as documented [here](https://github.com/iMerica/dj-rest-auth/pull/110) and [here](https://github.com/pennersr/django-allauth/issues/1975#issuecomment-384075169). The code behind `user.delete()` tries to delete associated records from `socialaccount` tables that are only created when the app is added to `INSTALLED_APPS` and migrations are applied. Make sure to run migrations after switching to this branch!

### DIFF
--- a/backend/lettercraft/settings.py
+++ b/backend/lettercraft/settings.py
@@ -164,6 +164,10 @@ REST_AUTH = {
     "USER_DETAILS_SERIALIZER": "user.serializers.CustomUserDetailsSerializer",
 }
 
+# Explicitly sets the auto field to the Django 3.2+ default,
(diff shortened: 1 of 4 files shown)


lukavdplas (approved):
Weird that socialaccounts is required, but whatever. Looks good!

PR #158, opened 2024-12-10
Feature/historical figures

### LINKED ISSUE #124: Agent form: historical figures
Add a menu to edit the relationships between agents and historical figures in the agent form.

For individuals, this is a single-select menu; for groups, you should be able to select multiple historic figures.

XanderVertegaal:
Closes #124 

Adds a control for historical person(s) to the Agent form. If the agent is marked as a 'group', the user is able to select multiple historical people; if the agent is a 'person', only one selected historical person is allowed. If an agent is switched from 'group' to 'person' when more than one historical person is connected, the form is invalid and an error message appears under the isGroup form control.


frontend/src/app/data-entry/agent-form/agent-historical-person-form/agent-historical-person-form.component.html
```diff
@@ -0,0 +1,17 @@
+<fieldset *ngIf="combined$ | async as combined">
+    <legend>
+        {{ combined.isGroup ? "Historical persons" : "Historical person" }}
+    </legend>
+    <p class="form-text">
+        {{
+            combined.isGroup
+                ? "Pick one or more historical persons that are described by this agent in this source."
+                : "Pick an historical person that is described by this agent in this source."
```
lukavdplas, line 9: Perhaps it would be clearer to say "historical person(s) that this agents represents". (So this is not historical figures that the agent diegetically describes.)

frontend/src/app/data-entry/agent-form/agent-historical-person-form/agent-historical-person-form.component.ts
```diff
@@ -0,0 +1,166 @@
...
+        .pipe(map((result) => result.data.historicalPersons));
+
+    private historicalPersonOptions$: Observable<MultiselectOption[]> =
+        this.allHistoricalPersons$.pipe(
+            map((persons) =>
+                persons.map((person) => ({
+                    value: person.id,
+                    label: `${person.name} (${
+                        person.dateOfBirth?.displayDate ?? " ?"
+                    } – ${person.dateOfDeath?.displayDate ?? "? "})`,
```
lukavdplas, line 59: If the dates are not filled in, you'll get, for example, `Radegund (? - ?)`. Looks a bit odd; at first I thought this was an emoticon of some sort.

I don't think we'll have that many historical figures anyway, so it would be fine to just use the name here.

frontend/src/app/data-entry/agent-form/agent-historical-person-form/agent-historical-person-form.component.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    private isGroup$ = this.agent$.pipe(
+        map((agent) => agent?.isGroup ?? false),
+        share()
+    );
+
+    public combined$ = combineLatest([
+        this.historicalPersonOptions$,
+        this.isGroup$,
+    ]).pipe(map(([options, isGroup]) => ({ options, isGroup })));
```
lukavdplas, line 72: Can you think of a more descriptive name for this?

frontend/src/app/data-entry/agent-form/agent-identification-form/agent-identification-form.component.html
```diff
@@ -36,11 +36,12 @@
         <label class="form-label" for="input-isgroup">
             Is this agent a person or a group?
         </label>
-        <select id="input-isgroup" class="form-select"
+        <select id="input-isgroup" class="form-select" [class.is-invalid]="form.invalid"
             [formControl]="form.controls.isGroup">
             <option [ngValue]="false">Person</option>
             <option [ngValue]="true">Group</option>
         </select>
+        <p class="invalid-feedback">A single-person agent cannot be linked to more than one historical person.</p>
```
lukavdplas, line 44: This is checking whether the form as a whole is invalid, so the message will also show up when there is an issue with the name:

screenshot of form, showing an empty name field, where the message about groups is also displayed

frontend/src/app/data-entry/agent-form/validators/isGroupValidator.ts
lukavdplas, line 1: Files in the frontend normally use snake case in the filename, so `is-group-validator.ts` would be more consistent.

But I think you could also just save this as `.../agent-form/validators.ts`, instead of making a file for one function only.

lukavdplas (changes requested):
Nice!

There is one bug in the code (with the validation); the rest are optional suggestions. Feel free to merge when you've handled my comments.

frontend/src/app/data-entry/agent-form/agent-identification-form/agent-identification-form.component.html
XanderVertegaal: Right, good spot!

PR #159, opened 2024-12-18
Feature/location form expansion

### LINKED ISSUE #151: Expand Location form
Add M2M fields to the Location logging form for Settlements, Structures and Regions.

XanderVertegaal:
Closes #151 

Adds subforms for Regions, Settlements and Structures.

The researchers like icons, so I've added a few to distinguish political, ecclesiastical and geographic regions, for instance. Bootstrap Icons did not have any suitable ones, so I had to get these from FontAwesome (as single SVGs, not the whole library). If you don't like them, I can easily remove them again.

image

I could not remember whether loggers should be able to add new regions/structures/settlements, or whether is the responsibility of the researchers themselves. Currently, loggers can only select created ones.


frontend/src/app/data-entry/location-form/location-regions-form/location-regions-form.component.html
```diff
@@ -0,0 +1,12 @@
+<form>
+    <p class="form-text">
+        Pick one or more regions that contain the location represented here.
```
lukavdplas, line 3: Maybe this can be a bit more clear that this is (at least for now) about regions that are actually mentioned.

E.g. "Select regions by which the text identifies or contextualises the location."

Same comment for settlements and structures.

frontend/src/assets/icons/church.svg
lukavdplas, line 1: If we distribute the icons in our own repository, we also have to include a  copy of the Font Awesome Free licence. (This is one of the conditions in the licence.)

That said, I wouldn't mind just adding the npm package with all icons.

lukavdplas (changes requested):
Looks good! The licensing needs to be addressed in order to merge, though.

lukavdplas (approved):

PR #164, opened 2025-01-29
Feature/source permissions

### LINKED ISSUE #160: Groups / permissions for sources
Two issues:

- With more students assistants and sources added, making all sources available to all students becomes chaotic. Add a system so assistants can be added to groups, and get editing rights to the set of sources for their group.
- For the public interface, allow marking some sources as public (probably a boolean property).

lukavdplas:
Changes permissions for contributors to be based on groups (close #160)

- Adds a `ContributorGroup` model to track cohorts of contributors. (As discussed, I did not use Django's `Group` model here because managing student assistants and being able edit admin site permissions should be different privileges.) Each group has access to its own set of source texts.
- The `is_contributor` boolean field is removed: users are contributors if they are added to a contributor group.
- Mutations related to a source text now require that the user is a member of an appropriate contributor group. If not, the mutation request will be rejected.
- Some queries now have an optional parameter `editable` to view only data you have permission to edit. This is now applied in the *Data entry* page, so users only see the sources they can edit.

The checks for mutations are not very DRY and create a lot of bulk in the code. Perhaps I'll still look for some abstraction there.


XanderVertegaal (approved):
Very nice! Works like a charm. 👍 

I noticed that the error message is displayed in the console and does not show up in the frontend, but this is definitely a separate issue (and only triggered if users are removed from a ContributorGroup _while_ they are editing).

image

Issue #143, opened 2024-10-04
Form changes appear to reset when navigating back/forward


lukavdplas:

lukavdplas:
This is sort of fixed in the latest patch release. However, the issue still occurs if you:

- edit designators (changes appear correctly)
- navigate to the source overview
- navigate back to the form

This also occurs with some other fields, we noticed the same thing with the summary field.

XanderVertegaal:
This should be fixed with the latest release (0.8.0).

PR #165, opened 2025-02-14
Fix: cache issue upon navigation

XanderVertegaal:
This is a proposed fix for #143. The problem described there is probably also there for the other forms, but I wanted to make sure that this solution is approved before I apply it to other parts of the application.

Whether we use `cache.evict()` for multiple fields or `refetchQueries` as I have done here is inconsequential (this is a bit shorter). The difficulty lies in avoiding multiple mutations: after the cache is evicted, `form.patchValue()` is called to set the new values, which always seems to trigger `valueChanges`, leading to another mutation. I tried adding `emitEvent: false` and `onlySelf: true`, but to no avail. My solution now is to prevent a second mutation with `distinctUntilChanged()`. It's not pretty, but it works. :)

### DIFF
--- a/frontend/src/app/data-entry/episode-form/episode-contents-form/episode-contents-form.component.ts
+++ b/frontend/src/app/data-entry/episode-form/episode-contents-form/episode-contents-form.component.ts
@@ -10,18 +10,20 @@ import {
 } from "generated/graphql";
 import {
     debounceTime,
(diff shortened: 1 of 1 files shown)


lukavdplas (approved):
Using `distinctUntilChanged` makes sense here, so this looks fine.

PR #170, opened 2025-02-26
Add "editable" property to graphql types

lukavdplas:
Adds a property `editable` to several graphql types to check whether the user is allowed to edit the object.


XanderVertegaal (approved):
Excellent, I have nothing to add here 👍

PR #185, opened 2025-03-31
Feature/gift letter categories

### LINKED ISSUE #184: Categories on Letters and Gifts
The form fields for categories still need to be implemented in the Gift and Letter forms.

Image

XanderVertegaal:
Closes #184 

This PR:
- homogenizes the GiftDescription and LetterDescription models. One used to inherit `Named` and the other did not.

- simplifies the `categories` field on both models: the through model (with unnecessary fields `source_mention` etc.) has been removed: `GiftDescription` and `LetterDescription` are now linked to `GiftCategory` and `LetterCategory` directly.

- Replaces the `HistoricalPersonSelect`, `LabelSelect` and `SubspaceSelect` components with a new SingleMultiSelect component (the name is not great). A lot of duplicated code is now out of the window. We now only use one component to handle all of these label/category/simple foreign key relations, and they have the same look/feel as well.

image


frontend/src/app/data-entry/shared/single-multi-select/single-multi-select.component.html
```diff
@@ -0,0 +1,29 @@
...
+                (click)="removeItem(item.value)"
+            >
+                <lc-icon [icon]="actionIcons.cancel" />
+            </button>
+        </li>
+    </ul>
+</div>
+
+<form *ngIf="multiple === true || control.value.length !== 1">
+    <lc-multiselect
```
XanderVertegaal, line 23: Since this SingleMultiSelectComponent is now the only component that uses `<lc-multiselect />` I considered merging the two and getting rid of the latter. In the end I decided against it because I liked how the inner component takes care of the `ngbDropdown` directives and the setting/removing of items from the dropdown list, while the outer component contains outward facing logic aimed at displaying selected options.

lukavdplas (approved):

PR #215, opened 2025-08-04
Feature/long entity lists

### LINKED ISSUE #187: Handle long lists in Add Entity dropdowns
The list of (especially) Agents can grow very large to the point where it becomes unwieldy. We should make the list scrollable (and limit it in size), and add a search bar so the user can find the agents they are looking for efficiently.

XanderVertegaal:
Closes #187 by:
- moving the "Create new ..." option to the top;
- sorting the available entities alphabetically.

I've added an extra commit that adds a search bar, but I'm not sure whether this is a good idea. I checked the list of [W3C APG patterns](https://www.w3.org/WAI/ARIA/apg/patterns/) but I cannot find a good comparandum. (The editable combobox comes close, but isn't quite the same.) I am happy to drop it if you don't like it. If we do keep it, we should decide on the best way to make this widget more accessible.

### DIFF
--- a/frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.html
+++ b/frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.html
@@ -30,7 +30,11 @@
                     Add {{entityName}}
                 </button>
                 <div ngbDropdownMenu [attr.aria-labelledby]="dropdownToggleID">
-                    <button ngbDropdownItem (click)="createEntity$.next()">
+                    <input type="text" class="form-control search-bar"
+                        placeholder="Search for {{entityName}}s"
+                        [formControl]="searchControl"
+                        [attr.aria-label]="'Search for ' + entityName + 's'" />
+                    <button *ngIf="!searchControl.value" ngbDropdownItem (click)="createEntity$.next()">
                         <lc-icon [icon]="actionIcons.create" />
                         Create new {{entityName}}
                     </button>
@@ -38,6 +42,9 @@
                         ngbDropdownItem (click)="addEntity$.next(entity.id); dropdownToggle.focus()">
                         {{entity.name}}
                     </button>
+                    <ng-container *ngIf="searchControl.value && (availableEntities$ | async)?.length === 0">
+                        <button ngbDropdownItem disabled>No results found.</button>
+                    </ng-container>
                 </div>
             </div>
         </div>
--- a/frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.html
+++ b/frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.html
@@ -30,13 +30,13 @@
                     Add {{entityName}}
                 </button>
                 <div ngbDropdownMenu [attr.aria-labelledby]="dropdownToggleID">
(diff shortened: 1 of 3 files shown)


frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.html
```diff
@@ -30,14 +30,21 @@
                     Add {{entityName}}
                 </button>
                 <div ngbDropdownMenu [attr.aria-labelledby]="dropdownToggleID">
+                    <input type="text" class="form-control search-bar"
+                        placeholder="Search for {{entityName}}s"
+                        [formControl]="searchControl"
+                        [attr.aria-label]="'Search for ' + entityName + 's'" />
```
lukavdplas, line 36: Generally, for something like a dropdown-with-search you can look at the [combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/), but this has somewhat different interactions.

That said, bootstrap dropdowns are meant to be flexible and function more like a generic [disclosure pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/). As long as you don't add more specific semantics like `listbox`, the menu can include other elements. (Provided the semantics of these elements are clear, of course.)

I have a few comments on this:

- The `<input>` element does not have the `ngbDropdownItem` directive; I expect it is currently unreachable via keyboard.
- This input is search-as-you-type, but if you use a screen reader, this isn't apparent. That said, I don't think you can solve this without creating annoying clutter, so I would just leave this as-is.
- It makes sense if the search input is immediately followed by the result (especially when using keyboard navigation), but now there is an unrelated button between them. Move the create new button be above the search control.
- Because this menu is getting a bit cluttered, I would recommend using [headers](https://getbootstrap.com/docs/5.3/components/dropdowns/#headers) or a [divider](https://getbootstrap.com/docs/5.3/components/dropdowns/#dividers) to distinguish between the "new" and "existing" sections.

For a more lengthy discussion of dynamic search, I would recommend [this blog post](https://www.scottohara.me/blog/2022/02/05/dynamic-results.html).

frontend/src/app/data-entry/episode-form/episode-entities-form/episode-entities-form.component.html
```diff
@@ -30,14 +30,21 @@
...
                         ngbDropdownItem (click)="addEntity$.next(entity.id); dropdownToggle.focus()">
                         {{entity.name}}
                     </button>
-                    <button ngbDropdownItem (click)="createEntity$.next()">
-                        <lc-icon [icon]="actionIcons.create" />
-                        Create new {{entityName}}
-                    </button>
+                    <ng-container *ngIf="searchControl.value && (availableEntities$ | async)?.length === 0">
+                        <button ngbDropdownItem disabled>No results found.</button>
+                    </ng-container>
```
lukavdplas, line 47: This is not a button; you could add `role="presentation"` or something, but I think you can just use [text](https://getbootstrap.com/docs/5.3/components/dropdowns/#text).

You can also put this message in a live region for accessibility (see linked article above).

XanderVertegaal: Done and done!

PR #217, opened 2025-08-22
Feature/landing page

### LINKED ISSUE #204: Landing page
See the PPT ("Landing Page version 0.1") for inspiration.

- Single introductory sentence

- Slightly longer description

- 7 entry points:
  - Database principles
  - Search whole database
  - Search per text
  - Terminology
  - Lettercraft stories (#195)
  - Team (#182)

The team liked [the landing page for the Social Bodies project](https://socialbodies.bham.ac.uk/), but felt a bit lost at the bottom of the page. (No obvious links to other pages.

XanderVertegaal:
Closes #204 

Restyles the landing page and adds a theme file `lc_theme.scss` that overrides a couple of Bootstrap variables to give the website an overall medieval feel. As a result, the data-entry/browsing interfaces are automatically restyled as well, but we might have to tweak it here and there to

When approved, this branch doesn't have to be merged into develop right away. We can also show it to the researchers and/or deploy to ACC first to see whether they like it.


frontend/src/app/home/home.component.html
```diff
@@ -1,4 +1,134 @@
...
+                    </div>
+                </div>
+            </div>
+        </div>
+    </header>
 
+    <section class="banner bg-primary text-secondary p-5 mb-5 lh-lg border-y-gold">
+        <h2 class="text-center mb-4">Welcome!</h2>
+
+        <div class="banner-content">
```
lukavdplas, line 25: Because this is a longer piece of text, I would use the gold colour for the header but make the body text white or ecru, for more comfortable reading.

frontend/src/lc_theme.scss
```diff
@@ -0,0 +1,72 @@
+// Variables
+$lc_joinquil: #ffce1f;
+$lc_indigo: #003d66;
```
lukavdplas, line 3: Very nitpicky but I would call this "royal blue" rather than "indigo".

frontend/src/lc_theme.scss
```diff
@@ -0,0 +1,72 @@
...
+    border-bottom: 2px solid $lc_joinquil;
+}
+
+.border-top-gold {
+    border-top: 2px solid $lc_joinquil;
+}
+
+.border-bottom-gold {
+    border-bottom: 2px solid $lc_joinquil;
+}
```
lukavdplas, line 63: You should be able to get all of these styles from bootstrap's [border utilities](https://getbootstrap.com/docs/5.3/utilities/borders/#border), e.g. `border-top-gold` is equivalent to `border-top border-secondary border-2`. (Which has the advantage of not creating a combinatoric expansion of classes.)

frontend/src/app/core/menu/menu.component.html
```diff
@@ -1,19 +1,19 @@
-<nav class="navbar navbar-expand-lg bg-dark" aria-label="main navigation"
+<nav class="navbar navbar-expand-lg bg-primary border-bottom-gold" aria-label="main navigation"
     data-bs-theme="dark">
-    <div class="container-fluid">
-        <div class="navbar-brand">
-            Lettercraft & Epistolary Performance in Medieval Europe
+    <div class="container-fluid d-flex px-3 w-100">
+        <div class="navbar-brand h4 pb-0 text-secondary">
+            Lettercraft
```
lukavdplas, line 5: I agree that the long name was an issue. If we do want the full name here, we could also use a slightly bigger site header with full name and the nav links on a row below that.

frontend/src/app/data-entry/agent-form/agent-form.component.html
```diff
@@ -13,7 +13,7 @@
         <h1 class="mb-4">
             <lc-icon [icon]="agentIcon(agent)" />
             {{ agent.name }}
-            <span class="text-secondary"> ({{ agent.source.name }}) </span>
+            <span class="text-primary"> ({{ agent.source.name }}) </span>
```
lukavdplas, line 16: In the Bootstrap system, `secondary` is normally a more subtle colour than `primary`, so picking bright yellow causes a lot of mismatched styles like this one. Changing it to `text-primary` isn't really appropriate here; it makes the source stand out rather than fade away. (But there is now a "lexical gap" for a subtle colour.)

Perhaps you could override the `$theme-colors` sass map to add gold as `"accent"`, and make `"secondary"` gray or taupe. (Not sure how bootstrap utilities work here.)

(Also, because the primary colour is very dark, the contrast between black and primary text is very low, so I would avoid putting these next to each other in a single line.)

frontend/src/index.html
```diff
@@ -1,14 +1,24 @@
...
+        <link rel="preconnect" href="https://fonts.googleapis.com" />
+        <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+        <link
+            href="https://fonts.googleapis.com/css2?family=Cinzel:wght@400;500;600&display=swap"
+            rel="stylesheet"
+        />
+        <link
+            href="https://fonts.googleapis.com/css2?family=Crimson+Text:ital,wght@0,400;0,500;0,600;1,400&display=swap"
+            rel="stylesheet"
+        />
```
lukavdplas, line 19: I think it's generally recommended that (if possible) you include any fonts you want to use and do not fetch them from an API.

lukavdplas (approved):
Looks good! Added some minor comments.

frontend/src/app/core/menu/menu.component.html
XanderVertegaal: Yes, that could work 👌 Let's see what the researchers think. If they are keen on having the full title in the header, we can offer this as a solution. (I myself am not too fussed about it.)

frontend/src/lc_theme.scss
XanderVertegaal: My source for the naming of these colours (I could never have thought about 'jonquil' myself) is https://coolors.co/003d66-ffce1f-f1e9d2-faf8ef-722f37, but I doubt these names are standardised.

frontend/src/app/data-entry/agent-form/agent-form.component.html
XanderVertegaal: I've mapped `secondary` to a dark gray [("Jet")](https://imagecolorpicker.com/color-code/3e3b3f) and now use `accent` for the bright yellow "Jonquil". This makes a lot more sense than before! :)

PR #218, opened 2025-08-26
Bugfix/permissions

lukavdplas:
- Fixes exploit to view private sources (and related objects) in graphql API
- Adjusts some tests that were (inadvertently) using this exploit and making unauthorised requests
- Adds extra tests for queries


backend/event/queries.py
```diff
@@ -87,8 +87,9 @@ def resolve_episode_entity_link(
         entity: str,
         episode: str,
         entity_type: Entity,
-    ) -> EpisodeEntityLink:
+    ) -> Optional[EpisodeEntityLink]:
```
XanderVertegaal, line 90:
```suggestion
    ) -> Optional[EpisodeEntity]:
```

... right?

backend/user/permissions.py
```diff
@@ -8,23 +8,48 @@
...
 
 
-def editable_sources(
-    user: Union[User, AnonymousUser, None], sources: QuerySet[Source] = Source.objects
-) -> QuerySet[Source]:
-    if not user or user.is_anonymous:
-        return sources.none()
+type MaybeUser = Union[User, AnonymousUser, None]
 
+def editable_condition(user: MaybeUser):
```
XanderVertegaal, line 17:
```suggestion
def editable_condition(user: MaybeUser) -> Q:
```

backend/user/permissions.py
```diff
@@ -8,23 +8,48 @@
 SOURCE_NOT_PERMITTED_MSG = (
     "Mutation affects a source text for which you do not have contributor access"
 )
+MATCH_NONE = Q(pk__in=[])
+MATCH_ALL  = Q(pk__isnull=False)
```
XanderVertegaal, line 12: Nice!

backend/user/permissions.py
```diff
@@ -8,23 +8,48 @@
...
+
+    return is_public | editable_condition(user)
+
+
+def visible_sources(user: Union[User, AnonymousUser]) -> QuerySet[Source]:
+    return Source.objects.filter(visible_condition(user))
+
+
+def can_view_source(user: MaybeUser, source: Source) -> bool:
+    return visible_sources(user).contains(source)
```
XanderVertegaal, line 55: `user` is of type `MaybeUser` so it can technically be `None`. I'm getting a type warning here.

backend/user/permissions.py
```diff
@@ -8,23 +8,48 @@
...
+def can_edit_source(user: MaybeUser, source: Source) -> bool:
     """
     Whether a user is allowed to edit a source
     """
 
-    return source in editable_sources(user)
+    return editable_sources(user).contains(source)
+
+
+def visible_condition(user: MaybeUser):
```
XanderVertegaal, line 41:
```suggestion
def visible_condition(user: MaybeUser) -> Q:
```

backend/event/types/EpisodeType.py
XanderVertegaal, line 1: The `resolve_editable()` and `get_queryset()` methods can be removed in this class as well, right?

backend/core/types/EntityDescriptionType.py
```diff
@@ -30,13 +30,27 @@ class Meta:
...
+
+
     @staticmethod
     def resolve_contributors(
         parent: EntityDescription, info: ResolveInfo
     ) -> QuerySet[User]:
         return parent.contributors.all()
 
 
+    @staticmethod
```
XanderVertegaal, line 48: Since you are moving the field resolver here (smart move!), why not move the field `editable` here as well? Then you can remove the `editable` fields on all individual models that implement `EntityDescriptionType`.

XanderVertegaal (approved):
Great! I'm glad this DANGEROUS exploit is fixed :P In all seriousness: I appreciate the tests and the clear readable way in which you implemented this.

I've added some comments to make everything a tiny bit more DRY.

PR #225, opened 2025-10-16
Add blog posts

lukavdplas:
Adds a "case studies" section to replace the [blog](https://lettercraft.sites.uu.nl/category/blog/).

Adds (somewhat barebones) frontend pages and forms in the Django admin to create blog posts. The content is entered as HTML; the idea is that we would do the actual admin and formatting.

### DIFF
--- a/frontend/src/app/case-studies/case-study-view/case-study-view.component.html
+++ b/frontend/src/app/case-studies/case-study-view/case-study-view.component.html
@@ -16,12 +16,13 @@
             <div class="card-body">
                 <p class="card-text">
                     This is a case study based on
-                    <span *ngFor="let source of caseStudy.sources">
+                    <span *ngFor="let source of caseStudy.sources; last as last">
                         <a [routerLink]="['/data/sources', source.id]"
                             class="icon-link ms-1">
                             <lc-icon [icon]="dataIcons.source" />
                             {{source.name}}
                         </a>
+                        <span *ngIf="!last">, </span>
                     </span>
                 </p>
             </div>


backend/case_study/models.py
```diff
@@ -1,23 +1,48 @@
...
+
+    content = models.TextField(
+        blank=True,
+        help_text='Article content (in HTML)',
+    )
+
+    sources = models.ManyToManyField(
+        to=Source,
+        blank=True,
+        help_text='Source texts that are the focus of the case study',
```
XanderVertegaal:
```suggestion
        help_text='Source texts that are the focus of the case study',
        related_name='case_studies',
```

backend/case_study/models.py
```diff
@@ -1,23 +1,48 @@
...
+    )
+
+    episodes = models.ManyToManyField(
+        to=Episode,
         blank=True,
-        help_text="Series involved in this case study",
+        help_text='Epistolary episodes that are the focus of the case study'
```
XanderVertegaal:
```suggestion
        help_text='Epistolary episodes that are the focus of the case study',
        related_name='case_studies',
```

frontend/src/app/case-studies/case-study-view/footnotes/footnotes.directive.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Directive to fix hash links for footnotes
+ */
+@Directive({
+    selector: '[lcFootnotes]',
+    standalone: false,
+})
+export class FootnotesDirective {
+    constructor(
+        private el: ElementRef,
```
XanderVertegaal:
```suggestion
        private el: ElementRef<HTMLElement>,
```

You will then see that `findElement` has a type error: `querySelector` does not return `undefined` if it can't find anything, but `null`.

frontend/src/app/case-studies/case-study-view/case-study-view.component.html
```diff
@@ -0,0 +1,35 @@
...
+
+        <p class="text-secondary">
+            {{caseStudy.date | date}}
+        </p>
+
+        <div class="card bg-light">
+            <div class="card-body">
+                <p class="card-text">
+                    This is a case study based on
+                    <span *ngFor="let source of caseStudy.sources">
```
XanderVertegaal: This works, but multiple sources are now squished together without a comma or space. Perhaps this is another good place for `Intl.ListFormat` 😍

XanderVertegaal (approved):
Neat! I have a few minor comments to make it even cleaner. If you have time, it would also be nice to have a couple of case studies generated by he create_dev_dataset script, but feel free to leave this for another time.

frontend/src/app/case-studies/case-study-view/case-study-view.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    authorNames(caseStudy: CaseStudy): string {
+        const names = caseStudy.authors.map(author => author.fullName);
+        const formatter = new Intl.ListFormat('en');
+        return formatter.format(names);
+    }
+
+    sanitizedContent(caseStudy: CaseStudy) {
+        return this.sanitizer.bypassSecurityTrustHtml(caseStudy.content);
```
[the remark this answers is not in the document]
lukavdplas: It's needed because the wordpress HTML uses the `name` attribute to identify footnotes, which is not considered safe. (You can replace `name` instance with `id`, which is not sanitised, but it's easy to miss a few.)

But sanitising HTML is typically for displaying content by users. Since we are writing the HTML here, I would say this is an acceptable case for bypassing the sanitiser.

PR #235, opened 2025-11-18
Feature/pagination ellipses

### LINKED ISSUE #232: Use ellipsis in pagination component
The pagination for sources/episodes lists every page, but it should use an ellipsis if there are many pages.

Screenshot:

screenshot of page selector, listing pages 1 to 29

lukavdplas:
- Pagination uses ellipses (close #232)
- Also includes a page selector at the bottom of the page

### DIFF
--- a/frontend/src/app/data/shared/paginator/paginator.component.ts
+++ b/frontend/src/app/data/shared/paginator/paginator.component.ts
@@ -1,7 +1,6 @@
 import { Component, Input } from "@angular/core";
 import { PageResult } from "../../utils/pagination";
 
-let nextID = 0;
 
 @Component({
     templateUrl: './paginator.component.html',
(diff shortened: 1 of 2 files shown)


frontend/src/app/data/shared/paginator/paginator.component.ts
```diff
@@ -1,11 +1,17 @@
 import { Component, Input } from "@angular/core";
 import { PageResult } from "../../utils/pagination";
 
+let nextID = 0;
```
XanderVertegaal: Neat way to easily create unique IDs for each instance of this component. 👌!

XanderVertegaal (approved):
Code-wise, this is fine. Having a second paginator at the bottom of the page and the ellipsis are definitely improvements. I'm just not sure I like the placement. The right-alignment (due to `.ms-auto`) leaves a big gap on the left side of the page.

Without pretending that this would be an improvement: remove the `.ms-auto` (or replace it with `.mx-auto`) and add `.visually-hidden` to the `<p>` with the label ID. I would say it is pretty clear that this control changes pages so the 'Page' label feels redundant, but I'd understand it if you left it in for accessibility reasons.

PR #236, opened 2025-11-18
Profile page

lukavdplas:
New features:

- Adds a public profile page for contributors and a page listing all contributors
- Users can edit their profile description and picture in the settings form

Implementation details:

- Adds a new `ContributorRole` model which can be used to assign public roles to users (e.g. "project lead", "research assistant", etc.). These do not affect the user's permissions, they are only for public communication.
- Users are only visible to others if they have a "public role" assigned by an admin.

## Required server changes

After the data migration, contributors need to be assigned a public role to remain visible.


frontend/src/app/case-studies/case-study-view/case-study-view.component.html
```diff
@@ -5,7 +5,11 @@
         <h1 class="mt-4">{{caseStudy.name}}</h1>
 
         <p class="lead">
-            By {{authorNames(caseStudy)}}
+            By
+            <ng-container *ngFor="let author of caseStudy.authors; last as last">
+                <a [routerLink]="['/contributors', author.id]" class="link-dark"
```
XanderVertegaal: The `authorNames()` method can go as it's otherwise not used.

If the list of authors is empty, I just see the word `By`. It should never happen, but maybe it's good to hide `<p class="lead">` or add something like 'unknown authors' in that case.

backend/user/models.py
```diff
@@ -37,6 +38,70 @@ def can_edit_source(self, source) -> bool:
...
+    def full_name(self) -> str:
+        if self.first_name and self.last_name:
+            return f"{self.first_name} {self.last_name}"
+        elif self.first_name or self.last_name:
+            return self.first_name or self.last_name
+        else:
+            return "nameless contributor"
+
+    @property
+    def public_role(self) -> 'ContributorRole':
```
XanderVertegaal: Naming this `public_role` is crucial. I was confused for a bit why authors were not shown in the frontend unless they had their role defined. Authors-without-a-role suddenly seemed uncredited for their contributions.

backend/user/views.py
```diff
@@ -1,15 +1,18 @@
+import os
```
XanderVertegaal: (Unused import.)

frontend/src/app/routing/routes.ts
```diff
@@ -86,6 +88,22 @@ const routes: Routes = [
             }
         ],
     },
+    {
+        path: 'contributors',
+        children: [
```
XanderVertegaal: Visiting a page for a non-existing ID gives a weird blank page, but maybe that's your punishment for trying to craft URLs yourself :P

frontend/src/app/services/auth.service.ts
```diff
@@ -1,13 +1,14 @@
 import { Injectable } from '@angular/core';
 import { takeUntilDestroyed, toSignal } from '@angular/core/rxjs-interop';
```
XanderVertegaal: Unused imports.

frontend/src/app/services/auth.service.ts
```diff
@@ -111,13 +106,34 @@ export class AuthService {
...
+            map(_.constant(null)),
+        );
+    }
+
+    /** Picture route for PUT/DELETE requests. Assumes the user is already authenticated. */
+    private pictureRoute$(): Observable<string> {
+        return this.currentUser$.pipe(
+            take(1),
+            filter(user => !!user),
+            map(user => user ? this.authRoute(`pictures/${user.id}/`) : ''), // this condition is trivially true
```
XanderVertegaal: What do you mean with 'trivially true'?

backend/user/queries.py
```diff
@@ -27,3 +29,11 @@ def resolve_user_descriptions(
         info: ResolveInfo,
     ) -> QuerySet[User]:
         return UserType.get_queryset(User.objects, info)
+
+    @staticmethod
+    def resolve_contributor_roles(
+        parent: None,
+        info: ResolveInfo,
+    ) -> QuerySet[User]:
```
XanderVertegaal:
```suggestion
    ) -> QuerySet[ContributorRole]:
```

PR #263, opened 2026-01-08
Search results fixes

### LINKED ISSUE #261: Lingering search results
If you search through the episodes and the query returns no results, the results that were loaded last are still visible

To reproduce:

- Go to `/data/episodes`
- Make a query with no results

This also happens if you make a query with at least one results, followed by a query with no results.

Screenshots:

screenshot with empty query screenshot with query 'phone' showing the same results, followed by a message saying there are no results

lukavdplas:
- Close #261 
- If there are no results, the message shows the search term from used in the request, rather than the current value of the control. (This matters if requests take a long time.)
- If the search request returns an error, the page shows an error message.


frontend/src/app/data/episode-list/episode-list.component.html
```diff
@@ -8,29 +8,36 @@ <h1 class="mb-4">Epistolary episodes</h1>
...
+            <lc-episode-preview
+                *ngFor="let episode of pageData.episodes"
+                [episode]="episode"
+            />
+        </ng-container>
+        <lc-paginator [pageResult]="pageResult" />
+    } @else {
+        <div class="mt-3 text-center">
+            <p class="text-muted">
+                No episodes found matching "{{ searchTerm$ | async }}."
```
XanderVertegaal, line 38:
```suggestion
                No episodes found matching "{{ searchTerm$ | async }}".
```

frontend/src/app/data/source-list/source-list.component.html
```diff
@@ -13,29 +13,38 @@ <h1 class="mb-4">Source texts</h1>
...
-                </small>
-            </div>
-            <p class="mb-1">
-                <ng-container *ngIf="source.medievalTitle">
-                    <em>{{ source.medievalTitle }}</em>.
-                </ng-container>
+    @if ((searchError$ | async); as error) {
+        <p class="alert alert-danger">
+            Something went wrong while fetching results.
+            @if (error) {
```
XanderVertegaal, line 19: (While you're at it, you may consider migrating the existing `*ngIf`s in this file to the new `@if` syntax for consistency.)

frontend/src/app/data/episode-list/episode-list.component.ts
```diff
@@ -26,6 +26,16 @@ export class EpisodeListComponent {
...
+
+    searchTerm$ = this.searchResult$.pipe(
+        map(result => result.searchTerm)
+    );
+
+    searchError$ = this.searchResult$.pipe(
+        map(result => result.error)
     );
 
     public collectionData$ = this.searchResult$.pipe(
```
XanderVertegaal, line 41: `collectionData$` is now best marked 'private'.

XanderVertegaal (approved):
Great!

frontend/src/app/data/episode-list/episode-list.component.html
lukavdplas: Either is correct, depending on your prescriptive style of choice (https://en.wikipedia.org/wiki/Quotation_marks_in_English#Order_of_punctuation)

XanderVertegaal: I know, but in this case it suggests that I searched for "Radegund." instead of "Radegund" (which is in my search bar), which confused me.

Issue #95, opened 2024-07-17
Data download


lukavdplas:
Visitors to the database should be able to download the research data. This is important for reusability.

We're not sure about a suitable format for the download yet. Thoughts:

A substantial portion of users won't prefer a very "technical" format. A word document in which they can read through the data and make notes may be ideal.

Of course, it is also good to offer a machine-readable format.

Normally I would suggests CSV but our data is so linked that this would be cumbersome to work with. Maybe a JSON document?

lukavdplas:
I think this is complete now.

PR #272, opened 2026-01-23
JSON export

lukavdplas:
Part of #95 

Adds a function to export data to JSON. You can use it via the command line.

Does not yet include a download interface for users. I wanted to test this first to get an idea of the required time, so I can see if we need async scheduling / email notifications / etc.

Also, the function will load the full data object in memory before making a JSON dump, which I think should not be a problem? Again, I thought I would some testing first.

### DIFF
--- a/backend/download/export_json.py
+++ b/backend/download/export_json.py
@@ -1,7 +1,7 @@
 from typing import Dict, List, Any
-from pathlib import Path
 import json
 import re
+from io import TextIOWrapper
 
 from django.db.models import QuerySet, Value, CharField
 from django.db.models.functions import Concat, JSONObject, NullIf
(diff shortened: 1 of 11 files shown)


backend/download/export_json.py
```diff
@@ -0,0 +1,158 @@
...
+def _id_expression(model_name: str, field: str = 'pk'):
+    '''
+    Expression to create ID annotation for queryset that includes the model name
+
+    Creates IDs like "agent/1" for an AgentDescription record.
+
+    Includes a check to return null if there is no PK value, which happens in ArrayAgg
+    when the list is empty.
+    '''
+    return NullIf(
```
XanderVertegaal, line 126: Clever!

XanderVertegaal (approved):
Nice work! Feel free to merge.

Just out of curiosity: did you have a particular usecase in mind  when you decided to represent IDs as "agent/16" instead of just "16"? Is it just to make the IDs themselves a bit more context-independent when you're reading the JSON?

lukavdplas:
> Just out of curiosity: did you have a particular usecase in mind  when you decided to represent IDs as "agent/16" instead of just "16"? Is it just to make the IDs themselves a bit more context-independent when you're reading the JSON?

Yes it's so the IDs are clearer in the JSON as a whole. As there can be object from multiple tables with id 16 otherwise.

PR #264, opened 2026-01-09
Feature/multiword search

lukavdplas:
close #262

Issue #262, opened 2026-01-08
Multi-term search


lukavdplas:
If you use a multi-word search string in the browse interface (e.g. `radegund bishop`), you search for the full string (so "the bishop writes to Radegund" would not match). The expected (and much more useful) behaviour would be to search for the individual terms (i.e. episodes containing `radegund` AND `bishop`).

PR #273, opened 2026-01-31
Feature: new browsing interface

XanderVertegaal:
This is a first working version of the browing interface prototype as demonstrated last meeting, allowing searching for text and labels (episode categories). Search results link to their respective detail pages. The component retrieves counts for all categories but only retrieves the actual data for the selected tab.

This does not have to be merged right away, but now we have at least a basis to (both) build upon :)

Several features are still missing but I chose not to add them now to avoid PR bloat.
- Encode the query in query parameters so users can visit detail pages and come back to their search results.
- Add a search bar to the label selector (There are only ~20 of them on production right now, so this is perhaps not really needed.)
- Make clear why Agents/Letters/Gifts/Locations are selected if a certain label is selected. It is because they are linked to a label through an episode, but this is not clear from the list of results itself.
- Re-add pagination to the result lists.


backend/search/queries.py
lukavdplas, line 1: Generally speaking, the `core` app functions as a package from which other packages import base classes/functions. So for a combined search like this, it would make more sense in the hierarchy to add a new django app (e.g. `search`) that imports from everything else.

backend/letter/types/GiftDescriptionType.py
```diff
@@ -1,14 +1,37 @@
...
+    search = CharFilter(method="search_gift_descriptions")
+    label_ids = CharInFilter(method="filter_by_labels")
+
+    def search_gift_descriptions(
+        self, queryset: QuerySet[GiftDescription], name: str, value: str
+    ) -> QuerySet[GiftDescription]:
+        """Filter gift descriptions by name or description."""
+        return queryset.filter(
+            Q(name__icontains=value) | Q(description__icontains=value)
+        )
```
lukavdplas: This (and the identical methods for agents, letters, and locations) only works for single-term queries, see https://github.com/CentreForDigitalHumanities/lettercraft/issues/262 - which was fixed by https://github.com/CentreForDigitalHumanities/lettercraft/pull/264/. That solution included a utility function you can import here.

frontend/src/app/data/browse/search-bar/search-bar.component.html
```diff
@@ -0,0 +1,21 @@
...
+    <button class="btn btn-primary search-button" type="submit">
+        @if (loading()) {
+        <div class="spinner-border spinner-border-sm me-2" role="status">
+            <span class="visually-hidden">Searching...</span>
+        </div>
+        } @else {
+        <lc-icon [icon]="actionIcons.search" class="me-2" aria-hidden="true" />
+        Search
+    }
+    </button>
```
lukavdplas: The placement this submit button in the containing form is odd:

screenshot showing (respectively) search text input, search button, and dropdown selector for labels

The submit button is now attached to the text input, but in reality it's also the submit button for the label selector. If you're only selecting a label, you would not realise that you need to hit "search" in the row above it.

Code-wise, it would also make more sense to remove the submit button from this component and add it to the `BrowseComponent` that defines the containing `<form>` element.

frontend/src/app/data/browse/browse.component.html
```diff
@@ -0,0 +1,90 @@
...
+                                    size="3x"
+                                />
+                                <p class="text-muted">
+                                    No results found in this category
+                                </p>
+                            </div>
+                            } @else {
+                            <div class="list-group">
+                                @for (item of tabItems; track item.id) {
+                                <lc-browse-list-item [listItem]="item" />
```
lukavdplas, line 89: For sources and episodes, I think it would be better if we continue using the existing source preview /episode preview components, as we've already workshopped those with the researchers. This new layout would remove a lot of information, which would feel like we're regressing on the feedback from last meeting (which was to add more information in the previews).

frontend/src/app/data/browse/browse-label-select/browse-label-select.component.html
lukavdplas, line 1: This component has a few keyboard accessibility issues:
- The remove button for individual labels cannot be used by keyboard
- Keyboard focus on the "remove all" button is barely visible
- Keyboard focus on the dropdown trigger button makes the "remove all" button disappear visually.

Also, the state of the menu items is only indicated in the CSS class, they should set `aria-pressed` as well.

frontend/src/app/data/browse/browse.component.ts
lukavdplas, line 1: A small reasoning error in this component: when you switch tabs, the form data also gets submitted, which is a weird side effect. So if you change something in the form, then switch to a different tab, the number in the tab can suddenly change. Which looks odd, since you would not expect the tabs to affect the results.

It can also give the appearance of "deception", as in the app claims 200 episodes match your search term, but then when want to inspect those, it goes "nevermind it's only 2".

frontend/src/app/data/browse/browse.component.ts
```diff
@@ -0,0 +1,207 @@
...
+                subText.push(`chapter ${episode.chapter}`);
+            }
+            if (episode.page) {
+                subText.push(`page ${episode.page}`);
+            }
+            return {
+                id: episode.id,
+                name: episode.name,
+                description: episode.summary,
+                subtext: subText.join(', '),
```
lukavdplas: This seems to introduce a leading comma:

screenshot of search item with text ',  book 1, chapter 2, page 3'

screenshot of search item with text ', page 52'

frontend/src/app/data/browse/search-bar/search-bar.component.ts
lukavdplas, line 1: It's not clear to me why this component exists, at it seems to be a wrapper around a standard text input. It looks like you could just save a lot of code by writing `<input type="text" [formControls]="form.controls.searchTerm">` directly in the `BrowseComponent`.

frontend/src/app/data/browse/search-bar/search-bar.component.html
```diff
@@ -0,0 +1,21 @@
...
+        class="form-control search-input"
+        name="search"
+        [value]="searchValue()"
+        (input)="onInputChange($event)"
+        placeholder="Search across all entities..."
+        aria-label="Search"
+    />
+    <button class="btn btn-primary search-button" type="submit">
+        @if (loading()) {
+        <div class="spinner-border spinner-border-sm me-2" role="status">
```
lukavdplas: Not a major issue, but for your information: live regions (like `role="status"`) report updates to their inner content to assistive technology. If you're adding/removing the element itself from the DOM (by wrapping it in an `@if`), making it a live region doesn't actually do anything.

frontend/src/app/data/browse/browse.component.html
```diff
@@ -0,0 +1,90 @@
...
+                                    class="text-muted mb-3"
+                                    size="3x"
+                                />
+                                <p class="text-muted">
+                                    No results found in this category
+                                </p>
+                            </div>
+                            } @else {
+                            <div class="list-group">
+                                @for (item of tabItems; track item.id) {
```
lukavdplas, line 88: This removes the paginator we added earlier, which is probably not a good thing?

frontend/src/app/data/browse/browse.component.scss
lukavdplas, line 1: I'm not sure what to make of the CSS in this PR. It's often overly verbose, e.g. by including rules that are already the default, or which could be handled more succinctly with a bootstrap class. There are also many rules that make micro-adjustments to the bootstrap layout which seem unnecessary.

For instance, the rules under the `.no-results` selector (lines 64-72) create this appearance:

screenshot of no results message

But if you remove these lines it looks like this, which is almost identical:

screenshot of no results message with slightly smaller font size

You know, what is the point of this?

lukavdplas (commented):
This was a lot of work, well done. That said, please try to avoid making huge PRs like this 😅  This one could have been split into two or three incremental PRs.

Since you made this as a rough draft, some of my comment here may have been in your own to-do list? But here you go.

XanderVertegaal:
I'm sorry about the size 😞  I'll try to keep PRs more succinct in the future.

backend/letter/types/GiftDescriptionType.py
XanderVertegaal: This looks great, but I cannot find the utility function. #264 was merged into [this branch](https://github.com/CentreForDigitalHumanities/lettercraft/tree/bugfix/search-results), but I don't think the latter has been merged into develop yet.

frontend/src/app/data/browse/browse.component.html
XanderVertegaal: No, the paginator should definitely be reintroduced. It's on the list of to-dos.

frontend/src/app/data/browse/browse.component.scss
XanderVertegaal: I've brought it back to 20 lines.

frontend/src/app/data/browse/browse.component.ts
XanderVertegaal: With the search button gone, this will be solved automatically.

backend/letter/types/GiftDescriptionType.py
lukavdplas: Wow, weird. It's merged now!

frontend/src/app/data/browse/browse.component.ts
XanderVertegaal: Fixed!

frontend/src/app/data/browse/browse-label-select/browse-label-select.component.html
XanderVertegaal: These should all be fixed now :)

XanderVertegaal:
I think I've gotten most of your comments, including the new multi-term search from #264. As discussed, I did not take over everything from that PR, since I had already started adapting BrowseListItem to have the same styling as the EpisodePreviewComponent. It can now be used for Sources, Agents, Gifts/Letters and Locations as well. 

Things to do:

- [The Letters/Gifts tab needs to be split](https://github.com/CentreForDigitalHumanities/lettercraft/issues/285)
- [Pagination needs to be readded](https://github.com/CentreForDigitalHumanities/lettercraft/issues/284)
- Perhaps remove the EpisodePreviewComponent or replace it with the BrowseListItem (as a way of code-deduplication)?

lukavdplas (approved):

PR #295, opened 2026-03-20
Split letters and gifts into separate tabs

### LINKED ISSUE #285: Split Letters/Gifts tab
Letters and Gifts are now lumped together in one tab in the BrowseComponent, but there is ample room for two tabs.

XanderVertegaal:
Closes #285.

### DIFF
--- a/backend/source/queries.py
+++ b/backend/source/queries.py
@@ -52,7 +52,7 @@ class SourceQueries(ObjectType):
 
     @staticmethod
     def resolve_sources(
-        root: None,
+        parent: None,
         info: ResolveInfo,
         editable=False,
         public_only=False,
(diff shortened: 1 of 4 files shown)


backend/source/queries.py
```diff
@@ -52,7 +52,7 @@ def resolve_source(
 
     @staticmethod
     def resolve_sources(
-        root: None,
+        parent: None,
```
XanderVertegaal, line 55: This may look weird, but it's the only resolver out of those for the browsable types that has 'root' as their first argument. The others all have 'parent'. This change allows me to put it in `common_params` in `queries.py`.

lukavdplas (approved):

PR #296, opened 2026-03-20
Feature/reintroduce paginator

### LINKED ISSUE #284: Add pagination to new browsing interface
The PaginatorComponent should be reintroduced to the source and episode list, which are now in the BrowseComponent, and its use should be extended to the Agents, Letters, Gifts and Location tabs as well.

XanderVertegaal:
Fixes #284


frontend/src/app/data/utils/pagination.ts
```diff
@@ -50,4 +51,12 @@ export class PageResult<Data> {
...
         return values.slice(start, end);
     }
+
+    private transform(data: Data): TransformedData {
+        if (this.transformer) {
+            return this.transformer(data);
+        }
+        // Safe cast: When no transformer is provided, TransformedData defaults to Data
+        return data as unknown as TransformedData;
+    }
```
lukavdplas, line 61: You could also consider using `_.identity` as the default value for `transformer`, that would make this method slightly more straightforward. Though `as unknown as TransformedData` would still be required as far as I know.

frontend/src/app/data/utils/pagination.ts
```diff
@@ -2,39 +2,40 @@ import { map, filter, switchMap, BehaviorSubject, combineLatest, Observable } fr
 import { ApolloQueryResult } from "@apollo/client/core";
 import { DestroyRef } from "@angular/core";
 import { takeUntilDestroyed } from "@angular/core/rxjs-interop";
+import { Paginated } from "../shared/paginator/paginator.component";
```
lukavdplas, line 5: Generally, I would expect component modules to import from utils modules, not the other way around. Perhaps the type definition should be in this module?

lukavdplas (approved):

PR #314, opened 2026-05-20
download button

lukavdplas:
Adds a frontend interface to download the project data as JSON.

### DIFF
--- a/backend/download/views.py
+++ b/backend/download/views.py
@@ -8,4 +8,7 @@ class DownloadJSONView(APIView):
     def get(self, request, *args, **kwargs):
         sources = Source.objects.filter(is_public=True)
         data = json_data(sources)
-        return JsonResponse(data)
+        filename = "lettercraft-data.json"
+        return JsonResponse(data, headers={
+            "Content-disposition": f"attachment; filename=\"{filename}\"",
+        })
--- a/backend/download/views.py
+++ b/backend/download/views.py
@@ -1,3 +1,11 @@
-from django.shortcuts import render
+from rest_framework.views import APIView
+from django.http.response import JsonResponse
+from source.models import Source
 
-# Create your views here.
+from download.export_json import json_data
+
+class DownloadJSONView(APIView):
+    def get(self, request, *args, **kwargs):
+        sources = Source.objects.filter(is_public=True)
+        data = json_data(sources)
+        return JsonResponse(data)
--- a/frontend/src/app/data/download/download.component.html
+++ b/frontend/src/app/data/download/download.component.html
@@ -1,0 +1,1 @@
+<a href="/api/download/download">Download data</a>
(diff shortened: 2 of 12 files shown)


frontend/src/app/shared/base-modal/base-modal.component.ts
```diff
@@ -8,6 +8,7 @@ import { Component, EventEmitter, Input, Output } from "@angular/core";
 })
 export class BaseModalComponent {
     @Input() title = "Lettercraft";
+    @Input() showFooter: boolean = true;
```
XanderVertegaal, line 11: The type is inferred so you don't have to indicate it here (although it doesn't hurt either).

frontend/src/app/shared/base-modal/base-modal.component.html
```diff
@@ -10,6 +10,6 @@
 <section class="modal-body">
     <ng-content select="[modal-content]" />
 </section>
-<footer class="modal-footer">
+<footer class="modal-footer" [hidden]="!showFooter">
```
XanderVertegaal, line 13: I am a bit scared to admit I didn't know about the global `hidden` attribute. Nice to see that there is a native way to hide any HTML element like this, but to keep it in the DOM.



Out of curiosity: is there a special reason why you went with `hidden` instead of `@if`/`*ngIf`?

frontend/src/app/data/download/download.component.html
```diff
@@ -0,0 +1,31 @@
+<button class="btn btn-primary" (click)="open()">
+    <lc-icon [icon]="icons.download" class="me-1" />
+    Download data
+</button>
+
+<ng-template #downloadModal>
+    <lc-base-modal (dismiss)="close()" [showFooter]="false">
+        <ng-container modal-header>
+            <h4 ngbAutofocus tabindex="-1">Download data</h4>
```
XanderVertegaal:
```suggestion
            <h4 ngbAutofocus tabindex="-1" class="mb-0">Download data</h4>
```

(The h4 is a bit above the center line now.)

backend/download/views.py
```diff
@@ -1,3 +1,14 @@
...
-# Create your views here.
+from download.export_json import json_data
+
+class DownloadJSONView(APIView):
+    def get(self, request, *args, **kwargs):
+        sources = Source.objects.filter(is_public=True)
+        data = json_data(sources)
+        filename = "lettercraft-data.json"
+        return JsonResponse(data, headers={
+            "Content-disposition": f"attachment; filename=\"{filename}\"",
```
XanderVertegaal, line 13: Neat use of Content-Disposition here!

XanderVertegaal (approved):
Great, this works very well! My only suggestion for improvement would be to show a small loading spinner (while the request is in flight) to show that the request is being processed, but this is not required for merging.

frontend/src/app/shared/base-modal/base-modal.component.html
lukavdplas: Content projection is not designed to be conditional. So you should either use something like this to hide the element (but still render it), or use template fragments to render it conditionally. See https://v20.angular.dev/guide/components/content-projection

PR #315, opened 2026-05-21
Feature/docx export

### LINKED ISSUE #95: Data download

lukavdplas:
Builds on #314 - adds the option to download the data as DOCX.

The docx contains almost all the data in the JSON export, except for fields that contain HTML data (source references, most notably). I want to add those too, but I'll pick that up in a separate PR.

Also adds the URL and date to the JSON download.

### DIFF
--- a/backend/download/tests/test_export_docx.py
+++ b/backend/download/tests/test_export_docx.py
@@ -1,9 +1,14 @@
 from source.models import Source
 from download.export_docx import save_docx
+from docx import Document
 
 
-def test_export_json(tmp_path, source: Source, episode, episode_2, episode_attribution):
+def test_export_docx(tmp_path, source: Source, episode, episode_2, episode_attribution):
     path = tmp_path / "data.docx"
     qs = Source.objects.filter(pk=source.pk)
     with open(path, "wb") as f:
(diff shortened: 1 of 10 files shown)


backend/download/tests/test_export_docx.py
```diff
@@ -0,0 +1,9 @@
+from source.models import Source
+from download.export_docx import save_docx
+
+
+def test_export_json(tmp_path, source: Source, episode, episode_2, episode_attribution):
```
XanderVertegaal: You probably forgot to rename this function 😄 Also: there are no assertions here (as in the corresponding _real_`test_export_json`). You may have intended this as a way to check that the function simply runs at all, but I'm just making sure you didn't forget :)

XanderVertegaal (approved):
Great!

PR #322, opened 2026-05-26
Feature/name sorting

lukavdplas:
I've been working on generating a citation file for the data - here is a related update to improve sorting for contributors.

- Some user lists in the interface are now sorted alphabetically, rather than by ID
- Adds a `last_name_prefix` database field so Dutch names can be sorted properly. Big day for me.


backend/user/migrations/0013_user_last_name_prefix.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('user', '0012_alter_userprofile_role'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='user',
+            name='last_name_prefix',
+            field=models.CharField(default='', help_text='Tussenvoegsel for Dutch localised names', max_length=64),
```
XanderVertegaal, line 16: I shiver at the thought of a 64 character tussenvoegsel.

backend/user/types/UserType.py
```diff
@@ -51,7 +52,7 @@ def sources_from_contributions(Model: Type[Model]):
             sources_from_contributions(SpaceDescription),
         )
 
-        return Source.objects.filter(id__in=source_ids)
+        return Source.objects.filter(id__in=source_ids, is_public=True)
```
XanderVertegaal, line 55: Good spot!

frontend/src/app/core/menu/user-menu/user-menu.component.spec.ts
```diff
@@ -24,7 +25,8 @@ const fakeAdminResponse: UserResponse = {
     username: "gandalf",
     email: "gandalf@istari.me",
     first_name: "Gandalf",
-    last_name: "The Grey",
+    last_name_prefix: 'the',
```
XanderVertegaal, line 28: 😂😂😂

XanderVertegaal (changes requested):
I respect your struggles as a last name prefix 'wielder'; please avoid creating struggles for poor non-wielders like me :P

`User.last_name_prefix` is a required field now, in both the frontend and Django Admin. In fact, it is the only required field. "Vincent Gogh" is not valid, but "van" is.

Please add `blank=True` to `last_name_prefix`. Then this is ready to be merged as far as I'm concerned! :)

PR #326, opened 2026-06-10
tabs in source page

lukavdplas:
Uses tabs layout on the source page so you can also see the agents/letters/etc.

Outfactors a `BrowseTabsComponent` from the `BrowseComponent` which is reused here.


backend/event/types/EpisodeSpaceType.py
```diff
@@ -1,19 +1,27 @@
...
     entity = NonNull(EntityDescription)
+    location = NonNull('space.types.SpaceDescriptionType.SpaceDescriptionType')
 
     class Meta:
         model = EpisodeSpace
         fields = [
             "id",
             "episode",
-            "space",
+            "location",
```
XanderVertegaal: `location` is a custom field, not a field on the EpisodeSpace model, so does not have to be listed in `Meta.fields`.

backend/event/types/EpisodeType.py
```diff
@@ -48,7 +48,7 @@ class Meta:
             "designators",
             "gifts",
             "letters",
-            "spaces",
+            "locations",
```
XanderVertegaal:
```suggestion
```

backend/search/test_queries.py
```diff
@@ -2,78 +2,74 @@ def test_search_sources_without_term(
...
-        "focus": "SOURCES",
         "term": "",
         "labelIds": [],
     }
     result = graphql_client.execute(
         query, variable_values=variables, context=anonymous_request
     )
 
-    assert result["data"]["search"]["sourceCount"] == 2
+    print(result)
```
XanderVertegaal: Rogue print statement!

frontend/src/app/data/browse/search-item/browse-list-item.component.spec.ts
```diff
@@ -9,6 +9,7 @@ describe('BrowseListItemComponent', () => {
 
     beforeEach(async () => {
         await TestBed.configureTestingModule({
+            declarations: [BrowseListItemComponent],
```
XanderVertegaal, line 12: This declaration is not required.

frontend/src/app/data/browse/browse.component.ts
```diff
@@ -85,16 +57,18 @@ export class BrowseComponent {
         shareReplay(1),
     );
 
+    searchTabs = [
```
XanderVertegaal: The difference between `searchTabs` in this component and `browseTabs` in the `SourceViewComponent` only differ in whether they include `SearchFocus.Sources` or not and will otherwise never change.

I suggest moving these tabs to the `BrowseTabsComponent` proper and filtering them there using the `hideSource` boolean input that is already there.

That way, you don't need to use `ngOnChanges`/`effect` in the `BrowseTabsComponent` to deal with the timing mismatch between prop initialization and rendering (and you don't need the added `tabs` input signal).

frontend/src/app/data/browse-tabs/browse-tabs.component.ts
```diff
@@ -0,0 +1,155 @@
...
+@Component({
+    selector: 'lc-browse-tabs',
+    standalone: false,
+    templateUrl: './browse-tabs.component.html',
+    styleUrl: './browse-tabs.component.scss'
+})
+export class BrowseTabsComponent {
+    tabs = input.required<SearchFocus[]>();
+    data = input.required<TabData | null>();
+    hideSource = input<boolean>(false);
```
XanderVertegaal, line 71:
```suggestion
    hideSource = input(false);
```

frontend/src/app/data/browse-tabs/browse-tabs.component.ts
```diff
@@ -0,0 +1,155 @@
...
+            [SearchFocus.Letters, data?.letters?.length],
+            [SearchFocus.Gifts, data?.gifts?.length],
+            [SearchFocus.Locations, data?.locations?.length],
+        ]);
+    });
+    loading = computed(() => !this.data());
+
+    filterItem = computed<(i: BrowseListItem) => BrowseListItem>(() =>
+        this.hideSource() ?
+        ((item: BrowseListItem) => _.omit(item, 'source') as any)
```
XanderVertegaal, line 101: Still not great, but more helpful (`source` is an optional property on all members of the union):

```suggestion
        ((item: BrowseListItem) => _.omit(item, 'source') as BrowseListItem)
```

frontend/src/app/data/source-view/source-view.component.ts
```diff
@@ -7,6 +7,7 @@ import { map, Observable, switchMap, filter } from 'rxjs';
 import { sourceBreadcrumbs } from '../utils/breadcrumbs';
 import { PageResult } from '../utils/pagination';
```
XanderVertegaal: Unused imports.

frontend/src/app/data/source-view/source-view.component.spec.ts
```diff
@@ -13,6 +15,8 @@ describe('SourceViewComponent', () => {
             declarations: [
                 SourceViewComponent,
                 ObjectPageHeaderComponent,
+                BrowseTabsComponent,
+                BrowseListItemComponent,
```
XanderVertegaal, line 19: These declarations appear to be unnecessary.

frontend/src/app/data/browse-tabs/browse-tabs.component.ts
```diff
@@ -0,0 +1,155 @@
...
+        private destroyRef: DestroyRef,
+        private sourcesPageQuery: BrowseSourcesPageGQL,
+        private episodesPageQuery: BrowseEpisodesPageGQL,
+        private agentsPageQuery: BrowseAgentsPageGQL,
+        private lettersPageQuery: BrowseLettersPageGQL,
+        private giftsPageQuery: BrowseGiftsPageGQL,
+        private locationsPageQuery: BrowseLocationsPageGQL,
+    ) {}
+
+    ngOnChanges(changes: SimpleChanges) {
```
XanderVertegaal: See my comment in `BrowseComponent` as to why this and `tabs` are probably not needed. 

If you still want to keep a side effect based on (input) signals like this one, [effect](https://angular.dev/api/core/effect) is probably more idiomatic; if you decide to keep `ngOnChanges` in particular, please add `implements OnChanges` to the class declaration.

XanderVertegaal (changes requested):
Works great! I like that you managed to replace aggregating individual `counts` in the backend with simply counting IDs.

Tiny request for change: the `BrowseTabsComponent` now receives `tabs` and a `hideSource` prop. I think it needs either but does not require both. (See my comment.)

PR #334, opened 2026-06-17
glossary

lukavdplas:
Adds a glossary page. Contents can be edited in the admin site.

**Changes to deployment:** 
- The `glossary` app must be added in Django settings.
- Researchers need to be given permission to view/edit glossary data in the admin


frontend/src/app/glossary/glossary.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export interface APIGlossaryReference {
+    id: number,
+    name: string,
+    category: number,
+    reference: string,
+}
+
+export interface GlosssaryItem {
```
XanderVertegaal: Glossssssssssary

```suggestion
export interface GlossaryItem {
```

XanderVertegaal (approved):
Works like a charm!

